Issue #3, opened 2016-07-15
Add iptables rule on startup


mgoodness:
Me again!

Would you be interested in a PR that enables kube2iam to add the necessary iptables rule?

jtblin:
Yes that would be interesting. Maybe behind a flag e.g. --iptables so that
it is opt-in.

Would that work when kube2iam is run as a docker container with --net=host?

mgoodness:
> Yes that would be interesting. Maybe behind a flag e.g. --iptables so that it is opt-in.

My thought/plan exactly.

> Would that work when kube2iam is run as a docker container with --net=host?

Yup. Also requires iptables installed in the image, and the pod must run with a privileged securityContext.

The code I have is appending the rule as expected, but traffic is not yet being routed to the kube2iam pod. Will submit the PR once I've figured that out.

dedalusj:
Don't know if it is your problem as well but I was setting this up a couple of days ago and I found that the iptables rule needs to be changed from interface `docker0` to `cbr0`.

When kubernetes starts on a node it restarts docker pointing it to the `cbr0` interface as bridge. It puzzled me for a while. Hope it helps.

mgoodness:
@dedalusj I suspect my issue is very similar, with Calico + flannel making things complicated.

mgoodness:
I ended up disabling Calico, as it uses iptables to perform DNAT of its own. Need to research more to see if & how it can play nicely with kube2iam. In the meantime, I have things working with flannel using the `docker0` interface.

PR #4, opened 2016-07-18
Add iptables rule

mgoodness:
Addresses #3.

I'm extremely new to Go, so please forgive any mistakes. No nit is too picky.


iptables/iptables.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"errors"
+	"os"
+
+	"github.com/coreos/go-iptables/iptables"
+)
+
+// AddRule adds the required rule to the host's nat table
+func AddRule(appPort, metadataAddress, hostInterface string) error {
+	hostIP := os.Getenv("HOST_IP")
```
jtblin: Is `HOST_IP` something set by Kubernetes automatically? I'd rather not mix flags and environment variables, so I would prefer to move this to a flag like the others, it would be trivial to use the env var to set the flag e.g. `---host-ip=$HOST_IP`.

If I were to mix env vars and flags, I'd probably go with [viper](https://github.com/spf13/viper) but this is beyond what should be done in this PR, so let's stick with flag for now.

README.md
```diff
@@ -142,10 +177,12 @@ and only pass the role name in the `iam.amazonaws.com/role` annotation, otherwis
 ```
 $ kube2iam --help
 Usage of kube2iam:
+      --add-rule                       Add iptables rule (requires HOST_IP environment variable)
```
jtblin: I'd rather call this flag `--iptables` or `--add-iptables-rule` if we want to be clearer.

README.md
```diff
@@ -119,18 +154,18 @@ kind: Pod
 metadata:
   name: aws-cli
   labels:
-	name: aws-cli
+    name: aws-cli
```
jtblin: I guess I should be adding a `.editorconfig` to avoid this kind of spacing issue ;) No biggies, no need to change anything.

glide.yaml
```diff
@@ -10,7 +10,7 @@ import:
 - package: github.com/karlseguin/ccache
 - package: github.com/spf13/pflag
 - package: k8s.io/kubernetes
-  version: v1.2.4
+  version: v1.3.0
```
jtblin: I usually prefer to spread additional changes to different PRs to avoid batching changes but no worries, killing two birds with one stone 👍

jtblin:
Thanks a ton for this PR 👍, looks pretty good to me except for 2 changes:

- [ ] use flag for `host_ip` instead of env var
- [ ] rename `add-rule` flag

Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM alpine:3.3
 
 RUN apk --no-cache add \
-    ca-certificates
+    ca-certificates iptables
```
jtblin: Please go to the line after `ca-certificates` and use `\` to separate each line.

iptables/iptables.go
mgoodness: I'm using the downward API to provide the pod's IP address as HOST_IP. Agree that using it in a flag is better for consistency. Adding `--host-ip`.

mgoodness:
@jtblin 

Any thoughts on the latest commit? No particular urgency, just want to make sure it doesn't slip through the cracks.

jtblin:
Done, published `0.1.0`. Thanks a ton for the addition! 👍

PR #7, opened 2016-08-19
Add a default-role flag

### LINKED ISSUE #6: Add default role option
* add a `--default-role` flag
* use this role if no annotation found on pod and flag is set

bharrisau:
Default role provides a fallback when annotation is missing.

closes #6

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -19,12 +19,13 @@ type Server struct {
 	APIToken        string
 	AppPort         string
 	BaseRoleARN     string
+	DefaultIAMRole  string
 	IAMRoleKey      string
 	MetadataAddress string
 	HostInterface   string
 	HostIP          string
-	Insecure        bool
 	AddIPTablesRule bool
+	Insecure        bool
 	Verbose         bool
 	Version         bool
 	iam             *iam
@@ -84,7 +85,7 @@ func (s *Server) securityCredentialsHandler(w http.ResponseWriter, r *http.Reque
 
 func (s *Server) roleHandler(w http.ResponseWriter, r *http.Request) {
 	remoteIP := parseRemoteAddr(r.RemoteAddr)
-	role, err := s.store.Get(remoteIP)
+	role, err := s.getRole(remoteIP)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusNotFound)
 		return
@@ -127,7 +128,7 @@ func (s *Server) Run(host, token string, insecure bool) error {
 		return err
 	}
 	s.k8s = k8s
-	s.store = newStore(s.IAMRoleKey)
+	s.store = newStore(s.IAMRoleKey, s.DefaultIAMRole)
 	s.k8s.watchForPods(s.store)
 	s.iam = newIAM(s.BaseRoleARN)
 	r := mux.NewRouter()
--- a/cmd/store.go
+++ b/cmd/store.go
@@ -4,14 +4,16 @@ import (
 	"fmt"
 	"sync"
 
+	log "github.com/Sirupsen/logrus"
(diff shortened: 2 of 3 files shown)


bharrisau:
This will be one of the first Go PRs I've submitted. And it took ages to get the build going. So please forgive any obvious issues.

cmd/store.go
```diff
@@ -9,9 +9,10 @@ import (
 
 // store implements the k8s framework ResourceEventHandler interface.
 type store struct {
-	iamRoleKey string
-	mutex      sync.RWMutex
-	rolesByIP  map[string]string
+	iamRoleKey  string
+	defaultRole string
```
jtblin: Nit: follow alphabetical order within same types i.e. move above `iamRoleKey`

cmd/store.go
```diff
@@ -20,6 +21,9 @@ func (s *store) Get(IP string) (string, error) {
 	defer s.mutex.RUnlock()
 	if role, ok := s.rolesByIP[IP]; ok {
 		return role, nil
+	} else if s.defaultRole != "" {
```
jtblin: `else` is not needed here as above it returns if it finds the role.

cmd/store.go
```diff
@@ -66,9 +70,10 @@ func (s *store) OnDelete(obj interface{}) {
 	}
 }
 
-func newStore(key string) *store {
+func newStore(key string, defRole string) *store {
```
jtblin: nit: rename to `defaultRole`

cmd/server.go
```diff
@@ -84,7 +85,7 @@ func (s *Server) securityCredentialsHandler(w http.ResponseWriter, r *http.Reque
 
 func (s *Server) roleHandler(w http.ResponseWriter, r *http.Request) {
 	remoteIP := parseRemoteAddr(r.RemoteAddr)
-	role, err := s.store.Get(remoteIP)
+	role, err := s.getRole(remoteIP)
```
jtblin: Huh, `getRole` doesn't seem to be defined anywhere. It doesn't look like it's needed.

cmd/server.go
```diff
@@ -147,6 +148,7 @@ func NewServer() *Server {
 	return &Server{
 		AppPort:         "8181",
 		IAMRoleKey:      "iam.amazonaws.com/role",
+		DefaultIAMRole:  "",
```
jtblin: Not needed, by default, go strings are empty strings.

cmd/store.go
```diff
@@ -66,9 +70,10 @@ func (s *store) OnDelete(obj interface{}) {
...
 	return &store{
-		iamRoleKey: key,
-		rolesByIP:  make(map[string]string),
+		iamRoleKey:  key,
```
jtblin: Nit: follow alphabetical order.

jtblin:
Thanks a ton for the PR. A few minor comments but otherwise should be good to go!

cmd/server.go
bharrisau: I made it the same as the credentialsHandler, using the `getRole` method that you have defined with backoff.

jtblin: Oh I see, sorry that's the problem with the partial view in the PR. Probably was a good idea to change here as well then 👍

bharrisau:
Fixed above issues. Did the 'alphabet reorder things' for `server.go`, and had to add the 'log' import in 'store.go'

jtblin:
Awesome, thanks a ton! 👍

bharrisau:
Will quickly update the README if you haven't merged this yet.

bharrisau:
Ha, too late.

jtblin:
Oops, good point. No worries I'll do that, I want to update the dependencies as well.

PR #10, opened 2016-09-14
expand iptables command to be more readable

rickypai:
use verbose flags and organize into different lines to make the command more readable.


jtblin:
thanks 👍

PR #16, opened 2016-10-05
Added namespace role option

whizard:
To prevent "anyone" from adding an annotation in their Pod rc to any role we added a flag to use a role based on the namespace name. This way only users who have rights to create a pod in a namespace could assume a role defined by the namespace name.

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -15,22 +15,23 @@ import (
 // Server encapsulates all of the parameters necessary for starting up
 // the server. These can either be set via command line or directly.
 type Server struct {
-	APIServer       string
-	APIToken        string
-	AppPort         string
-	BaseRoleARN     string
-	DefaultIAMRole  string
-	IAMRoleKey      string
-	MetadataAddress string
-	HostInterface   string
-	HostIP          string
-	AddIPTablesRule bool
-	Insecure        bool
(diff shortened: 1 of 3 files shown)


SleepyBrett:
I'm sitting down to write a version of this, and I had the following plan... well couple of potential plans.

Enabled only via a flag.
Undetermined yet as to where to do the mapping:
 - option one: store it in a confgmap in a protected namespace, so an admin would have to add the mappings of roles to namespaces.
- option two: add the roles allowed to be assumed to the namespace itself via annotation (in my current cluster that would me an admin would have to touch that)

As far as I can see there is no way to enable a 'self service' approach to users in a namespace without potentially compromising roles in other namespaces (unless you do a 'only one namespace can claim a role', but that leads to some restrictions I don't quite like).

SleepyBrett:
I opened this, same idea less magic: https://github.com/jtblin/kube2iam/pull/39

fiunchinho:
Should this PR be closed in favor of #39?

whizard:
I think there are other ways to do this via config maps. However, in the other implementation I would make sure only pods in a namespace (those that can access the namespace config) should have access to the specified roles. Meaning, I want assumed roles to only be namespace contained.

PR #18, opened 2016-10-12
Get rid of error swallowing & handle DeletedFinalStateUnknown

### LINKED ISSUE #17: Pod got role from another pod
I'm not sure how to debug this further so I'm looking for help. I deployed kube2iam via a DaemonSet with the `--iptables` option. What I saw was this:

```
$ kubectl exec -it ladder /bin/bash
$ curl http://169.254.169.254/latest/meta-data/iam/security-credentials/ && echo
commander_role
$ curl http://169.254.169.254/latest/meta-data/iam/security-credentials/ && echo
commander_role
$ <restart local kube2iam pod>
$ curl http://169.254.169.254/latest/meta-data/iam/security-credentials/ && echo
ladder_role
```

The right role for this pod was `ladder_role`. For some reason kube2iam kept returning a role for an unrelated pod.

mikekap:
May fix #17. I'm running this version and will report back if I see any funny business with iam.

### DIFF
--- a/cmd/store.go
+++ b/cmd/store.go
@@ -6,6 +6,7 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 	"k8s.io/kubernetes/pkg/api"
+	kcache "k8s.io/kubernetes/pkg/client/cache"
 )
 
 // store implements the k8s framework ResourceEventHandler interface.
@@ -23,8 +24,8 @@ func (s *store) Get(IP string) (string, error) {
 	if role, ok := s.rolesByIP[IP]; ok {
 		return role, nil
 	}
+	log.Warnf("Using fallback role for IP %s", IP)
 	if s.defaultRole != "" {
-		log.Warnf("Using fallback role for IP %s", IP)
 		return s.defaultRole, nil
 	}
 	return "", fmt.Errorf("Unable to find role for IP %s", IP)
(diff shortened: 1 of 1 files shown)


mikekap:
This has been in prod for a bit and I haven't seen any more IAM issues in the last week. I think this is safe to go in.

cmd/store.go
```diff
@@ -23,52 +24,46 @@ func (s *store) Get(IP string) (string, error) {
 	if role, ok := s.rolesByIP[IP]; ok {
 		return role, nil
 	}
+	log.Warnf("Using fallback role for IP %s", IP)
```
jtblin: I don't think we need to move this log, we don't know if we are going to fallback to the default role yet.

cmd/store.go
```diff
@@ -23,52 +24,46 @@ func (s *store) Get(IP string) (string, error) {
...
 // OnAdd is called when a pod is added.
 func (s *store) OnAdd(obj interface{}) {
-	if pod, ok := obj.(*api.Pod); ok {
-		if pod.Status.PodIP != "" {
-			if role, ok := pod.Annotations[s.iamRoleKey]; ok {
-				s.mutex.Lock()
-				s.rolesByIP[pod.Status.PodIP] = role
-				s.mutex.Unlock()
-			}
+	pod := obj.(*api.Pod)
```
jtblin: With this change, it will panic which I don't think is the right thing to do. I agree that swallowing the error is not great but I'd prefer to log the error rather than panicking. Looking at a similar issue in the kubernetes project, they decided to log the error instead, see https://github.com/kubernetes/kubernetes/pull/18042/files

cmd/store.go
```diff
@@ -23,52 +24,46 @@ func (s *store) Get(IP string) (string, error) {
...
+			s.mutex.Unlock()
 		}
 	}
 }
 
 // OnUpdate is called when a pod is modified.
 func (s *store) OnUpdate(oldObj, newObj interface{}) {
-	oldPod, okOld := oldObj.(*api.Pod)
-	newPod, okNew := newObj.(*api.Pod)
+	oldPod := oldObj.(*api.Pod)
```
jtblin: Same as above, let's log the error instead of panicking.

cmd/store.go
```diff
@@ -23,52 +24,46 @@ func (s *store) Get(IP string) (string, error) {
...
 func (s *store) OnDelete(obj interface{}) {
-	if pod, ok := obj.(*api.Pod); ok {
-		if pod.Status.PodIP != "" {
-			s.mutex.Lock()
-			delete(s.rolesByIP, pod.Status.PodIP)
-			s.mutex.Unlock()
-		}
+	pod, ok := obj.(*api.Pod)
+	if !ok {
+		pod = obj.(kcache.DeletedFinalStateUnknown).Obj.(*api.Pod)
```
jtblin: Thanks, seems to make sense given the comment on https://github.com/kubernetes/kubernetes/blob/8fd414537b5143ab039cb910590237cabf4af783/pkg/client/cache/controller.go#L146. Good find 👍

cmd/store.go
jtblin: It is not the case of an unknown message sent, store is just an implementation of a `cache.ResourceEventHandler`. The only reason why something unknown would be in the cache is if there is a problem in kubernetes internals, which is very similar to the issue I linked. Crashing is definitely not the way to handle cast errors IMO, we should log the issue and return.

jtblin: And to be clear, crashing accomplishes nothing regarding resyncing all the pods, the non pod is in kubernetes cache, not kube2iam cache. 

Actually I am not even sure it is an error if something else than a pod is in the cache, could be a different type of object.

mikekap: The interface is generic, but we explicitly only watch for pods: https://github.com/jtblin/kube2iam/blob/master/cmd/k8s.go#L31 .

I'll make a change to log errors instead of crashing, but I think I'll keep running it with the crash code. The `Informer` cache is a big unknown to me and this `DeletedFinalStateUnknown` is probably just the tip of the iceberg. This value doesn't come from kubernetes itself - it's part of the informer framework (i.e. the API). Crashing is definitely safer from a security perspective and I'd like to keep that in my cluster.

jtblin:
Thanks!

PR #21, opened 2016-10-18
Prepend iptables rule instead of appending it

shadone:
This fixes an issue in Kubernetes when the appended rule is never activated
because it is swalled by pre-existing rules.

Without this PR here is what the PREROUTING table look like:

```
core@ip-10-170-23-119 ~ $ sudo iptables -t nat -L PREROUTING --line-numbers -n
Chain PREROUTING (policy ACCEPT)
num  target     prot opt source               destination
1    KUBE-SERVICES  all  --  0.0.0.0/0            0.0.0.0/0            /* kubernetes service portals */
2    DOCKER     all  --  0.0.0.0/0            0.0.0.0/0            ADDRTYPE match dst-type LOCAL
3    DNAT       tcp  --  0.0.0.0/0            169.254.169.254      tcp dpt:80 to:10.170.23.119:8181

core@ip-10-170-23-119 ~ $ sudo iptables -t nat -L KUBE-SERVICES --line-numbers -n
Chain KUBE-SERVICES (2 references)
num  target     prot opt source               destination
1    KUBE-SVC-DKLLMOGNZZKRVIWM  tcp  --  0.0.0.0/0            192.168.64.58        /* jenkins/jenkins-discovery:slaves cluster IP */ tcp dpt:50000
2    KUBE-SVC-TPZT2TPWTBFEHRUX  tcp  --  0.0.0.0/0            192.168.64.58        /* jenkins/jenkins-discovery:master cluster IP */ tcp dpt:80
3    KUBE-SVC-NPX46M4PTMTKRN6Y  tcp  --  0.0.0.0/0            192.168.64.1         /* default/kubernetes:https cluster IP */ tcp dpt:443
4    KUBE-SVC-XGLOHA7QRQ3V22RZ  tcp  --  0.0.0.0/0            192.168.64.188       /* kube-system/kubernetes-dashboard: cluster IP */ tcp dpt:80
5    KUBE-SVC-HFYXROO3G4KITGSW  tcp  --  0.0.0.0/0            192.168.64.224       /* jenkins/jenkins-ui:ui cluster IP */ tcp dpt:443
6    KUBE-SVC-C6PFMPKKF2D63UAO  tcp  --  0.0.0.0/0            192.168.64.2         /* default/frontend:frontend cluster IP */ tcp dpt:80
7    KUBE-SVC-TCOU7JCQXEZGVUNU  udp  --  0.0.0.0/0            192.168.64.10        /* kube-system/kube-dns:dns cluster IP */ udp dpt:53
8    KUBE-SVC-ERIFXISQEP7F7OF4  tcp  --  0.0.0.0/0            192.168.64.10        /* kube-system/kube-dns:dns-tcp cluster IP */ tcp dpt:53
9    KUBE-SVC-BJM46V3U5RZHCFRZ  tcp  --  0.0.0.0/0            192.168.64.165       /* kube-system/heapster: cluster IP */ tcp dpt:80
10   KUBE-NODEPORTS  all  --  0.0.0.0/0            0.0.0.0/0            /* kubernetes service nodeports; NOTE: this must be the last rule in this chain */ ADDRTYPE match dst-type LOCAL

core@ip-10-170-23-119 ~ $ sudo iptables -t nat -L DOCKER --line-numbers -n
Chain DOCKER (2 references)
num  target     prot opt source               destination
1    RETURN     all  --  0.0.0.0/0            0.0.0.0/0
2    RETURN     all  --  0.0.0.0/0            0.0.0.0/0```

### DIFF
--- a/iptables/iptables.go
+++ b/iptables/iptables.go
@@ -6,6 +6,21 @@ import (
 	"github.com/coreos/go-iptables/iptables"
 )
 
+func iptablesInsertUnique(ipt *iptables.IPTables, table, chain string, pos int, rulespec ...string) error {
+	exists, err := ipt.Exists(table, chain, rulespec...)
+	if err != nil {
+		return err
+	}
+
+	if exists {
+		if err := ipt.Delete(table, chain, rulespec...); err != nil {
+			return err
+		}
+	}
+
+	return ipt.Insert(table, chain, pos, rulespec...)
+}
+
 // AddRule adds the required rule to the host's nat table
 func AddRule(appPort, metadataAddress, hostInterface, hostIP string) error {
 	if hostIP == "" {
@@ -17,8 +32,8 @@ func AddRule(appPort, metadataAddress, hostInterface, hostIP string) error {
 		return err
 	}
 
-	if err := ipt.AppendUnique(
-		"nat", "PREROUTING", "-p", "tcp", "-d", metadataAddress, "--dport", "80",
+	if err := iptablesInsertUnique(ipt,
+		"nat", "PREROUTING", 1, "-p", "tcp", "-d", metadataAddress, "--dport", "80",
 		"-j", "DNAT", "--to-destination", hostIP+":"+appPort, "-i", hostInterface,
 	); err != nil {
 		return err

### CODE REVIEW

File: iptables/iptables.go [hunk 1]
```diff
@@ -6,6 +6,19 @@ import (
 	"github.com/coreos/go-iptables/iptables"
 )
 
+func iptablesAppendUnique(ipt *iptables.IPTables, table, chain string, pos int, rulespec ...string) error {
```
mgoodness: Maybe a nit, but I recommend renaming this `iptablesInsertUnique`.

shadone: Doh, that is a mistake, I meant to call it `iptablesPrependUnique`. I can rename it to InsertUnique but then it'd make sense to pass the position as an argument as well.

What do you think?

mgoodness:
If the rule is found to exist already, would it make sense to delete it and then insert again? Seems like that's the only way to guarantee it will always be first in the chain.

shadone:
Deleting the rule if it exists is definitely a good idea. Coincidentally I've just been debugging an issue why the rule wasn't prepended correctly and it turns out after node restart kubernetes added its own rules again and kube2iam decided not to do anything because the DNAT rule already existsed. Unfortunately the rule was moved down the chain and essentially stopped working.

I've pushed an updated PR that deletes the rule if it exists and will be testing a custom build of kube2iam in our test cluster.

shadone:
It seems to be working fine!

For reference here is out k8s daemonset definition:
```
---
apiVersion: extensions/v1beta1
kind: DaemonSet
metadata:
  name: kube2iam
  labels:
    app: kube2iam
spec:
  template:
    metadata:
      labels:
        name: kube2iam
    spec:
      hostNetwork: true
      containers:
        - name: kube2iam
          image: our.private.registry.com/kube2iam:latest
          args:
            - "--base-role-arn=arn:aws:iam::1234567890:role/"
            - "--default-role=arn:aws:iam::1234567890:role/DenyEverything"
            - "--iptables=true"
            - "--host-ip=$(HOST_IP)"
            - "--host-interface=cni+" # match any network interfaces starting with 'cni' (e.g 'cni0')
            - "--verbose"
          env:
            - name: HOST_IP
              valueFrom:
                fieldRef:
                  fieldPath: status.podIP
          ports:
            - containerPort: 8181
              hostPort: 8181
              name: http
          securityContext:
            privileged: true
          livenessProbe:
            exec:
              command:
                - /bin/sh
                - -c
                - iptables -t nat -L PREROUTING -n -v --line-numbers | grep '^1 ' | grep 169.254.169.254
            initialDelaySeconds: 10
            timeoutSeconds: 1
```

shadone:
It seems that I have made a mistake in my testing (using the wrong --host-interface at first), the ordering of iptables rules doesn't expose any issues and this PR is not actually needed. Sorry for the noise! 🤕

jtblin:
No worries, thanks for following-up.

PR #22, opened 2016-10-19
Fixed sample iptables command

shadone:
The example command given in the docs doesn't work because the `--protocol` argument should be listed before `--dport` on the command line.


jtblin:
Thanks! 👍

PR #30, opened 2016-11-08
pre-fetch credentials, and refresh them before they expire.

marshallbrekka:
Background for the PR in issue #31


jtblin:
Thanks for the PR. I added my comment to the original issue, I don't think this solution is going to scale.

PR #33, opened 2016-11-29
Make backoff MaxInterval and MaxElapsedTime configurable

### LINKED ISSUE #28: Don't hang forever if there is no role defined for a pod
Currently a call to metadata from a pod without a role will hang forever because it retries to get the role from the IP indefinitely: https://github.com/jtblin/kube2iam/blob/master/cmd/server.go#L61-L66.

If the retry really is needed I think it would be better to limit it to X number of retries and then respond. Hanging the connection forever is not nice for the calling application.

I know this could be solved by setting a default role, but it should also work when no default role is defined IMO.

I wouldn't mind making a PR fixing this, but I would like your (@jtblin) opinion before I start, if you don't mind.

* Do we really need to retry or would it be ok to respond with a 404, this obviously gives a false positive if kube2iam just is too slow to recognize the pod annotation?
* If we need to retry, can we limit it to a finite number of retries?

mikkeloscar:
This makes the max interval and max elapsed time configurabel for the
exponential backoff used when getting a role based on source IP.

The defaults are still the same e.g. 1 minute for MaxInterval and 15
minutes for the MaxElapsedTime.

Fix #28

### DIFF
--- a/main.go
+++ b/main.go
@@ -2,14 +2,19 @@ package main
 
 import (
 	"runtime"
+	"time"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/spf13/pflag"
 
-	"github.com/cenk/backoff"
 	"github.com/jtblin/kube2iam/cmd"
 	"github.com/jtblin/kube2iam/iptables"
 	"github.com/jtblin/kube2iam/version"
+)
+
+const (
+	defaultMaxInterval    = 1 * time.Second
+	defaultMaxElapsedTime = 2 * time.Second
 )
 
 func main() {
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -49,6 +50,8 @@ func addFlags(s *cmd.Server, fs *pflag.FlagSet) {
 	fs.BoolVar(&s.AddIPTablesRule, "iptables", false, "Add iptables rule (also requires --host-ip)")
 	fs.StringVar(&s.HostInterface, "host-interface", "docker0", "Host interface for proxying AWS metadata")
 	fs.StringVar(&s.HostIP, "host-ip", s.HostIP, "IP address of host")
+	fs.DurationVar(&s.BackoffMaxInterval, "backoff-max-interval", backoff.DefaultMaxInterval, "Max interval for backoff when querying for role.")
```
jtblin: I think I'd rather change the defaults to some saner values. We probably don't want to retry for more than a few seconds actually, what do you think?

jtblin:
Thanks!

mikkeloscar:
Thanks, any chance you plan to cut a release with the latest changes? :)

jtblin:
Yep gonna do this today real soon :)

mikkeloscar:
Awesome!

jtblin:
Done, https://github.com/jtblin/kube2iam/releases/tag/0.3.0 and published image as well https://hub.docker.com/r/jtblin/kube2iam/tags/

Issue #37, opened 2016-12-15
Valid iptable Interface format with wildcards(+) are treated as wrong


alexouzounis:
Since https://github.com/jtblin/kube2iam/commit/d55bfe9e275279d7d5244da42134b402b7025354 we get:

```
2016-12-15T14:10:18.858624928Z time="2016-12-15T14:10:18Z" level=fatal msg="route ip+net: no such network interface" 
```

This is probably because we use `cali+` as the interface and the check implemented in the above commit is marking it as incorrect even though its valid.

alexouzounis:
reverted to `0.2.2` and everything works as expected

alexouzounis:
iptables support;
```
If the interface name ends in a "+", then any interface which begins with this name will match. If this option is omitted, any interface name will match.
```
which the above commit does not take into account.

Let me know if you need any more info

johnt337:
I had commented to your notes on the PR. Moving my notes here:

So here is my example with using the interface as cbr0
![image](https://cloud.githubusercontent.com/assets/1334469/21240764/ae1a9dbe-c2da-11e6-8017-54224d173d14.png)

I traced that back to https://golang.org/src/net/interface.go line 155:

```
138 // InterfaceByName returns the interface specified by name.
139	func InterfaceByName(name string) (*Interface, error) {
140		if name == "" {
141			return nil, &OpError{Op: "route", Net: "ip+net", Source: nil, Addr: nil, Err: errInvalidInterfaceName}
142		}
143		ift, err := interfaceTable(0)
144		if err != nil {
145			return nil, &OpError{Op: "route", Net: "ip+net", Source: nil, Addr: nil, Err: err}
146		}
147		if len(ift) != 0 {
148			zoneCache.update(ift)
149		}
150		for _, ifi := range ift {
151			if name == ifi.Name {
152				return &ifi, nil
153			}
154		}
155		return nil, &OpError{Op: "route", Net: "ip+net", Source: nil, Addr: nil, Err: errNoSuchInterface}
156	}
```

It appears that sometimes the interfaces are not ready (or available)

My cluster is running: `Server Version: version.Info{Major:"1", Minor:"4", GitVersion:"v1.4.6", GitCommit:"e569a27d02001e343cb68086bc06d47804f62af6", GitTreeState:"clean", BuildDate:"2016-11-12T05:16:27Z", GoVersion:"go1.6.3", Compiler:"gc", Platform:"linux/amd64"}`

[...]

PR #38, opened 2016-12-16
Ignore wildcard interface names in interface check

johnt337:
# Patch to ignore interface names like cali+

## What is the problem/feature?

As per #37 calico networks are passed to kube2iam with the 'cali+' notation.
https://github.com/jtblin/kube2iam/pull/34 did not take this into consideration.

## How did it get fixed/implemented?

- Added conditional to `CheckInterfaceExists` that returns `nil` if a _+_ is detected
- Added a test to support this condition

### DIFF
--- a/iptables/iptables.go
+++ b/iptables/iptables.go
@@ -3,6 +3,7 @@ package iptables
 import (
 	"errors"
 	"net"
+	"strings"
 
 	"github.com/coreos/go-iptables/iptables"
 )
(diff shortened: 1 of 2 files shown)


iptables/iptables.go
```diff
@@ -33,8 +34,14 @@ func AddRule(appPort, metadataAddress, hostInterface, hostIP string) error {
 	return nil
 }
 
-// CheckInterfaceExists - validates the interface passed exists for the given system
+// CheckInterfaceExists - validates the interface passed exists for the given system, ignores calico networks
```
jtblin: Nit: it could be other networks as well, it's really about ignoring wildcard characters so i'd make this comment and the one below more generic.

iptables/iptables.go
```diff
@@ -33,8 +34,14 @@ func AddRule(appPort, metadataAddress, hostInterface, hostIP string) error {
...
 func CheckInterfaceExists(hostInterface string) error {
+
+	if strings.Contains(hostInterface, "+") {
+		// calico networks ignored
```
johnt337: Changed from "calico networks ignored" to "wildcard networks ignored"

Fsero:
I'm also interested on this

pastjean:
Hey! Just wondering when you'd merge this. I'm currently running a custom version. (I could also help maintain)

jtblin:
Sorry for the lag, I was on annual leave. Thanks for the fix! 👍

PR #39, opened 2016-12-16
Namespace restriction

SleepyBrett:
This pull request implements a namespace restriction scheme. It's useful in multi-tennent clusters. Roles specified in pod annotations are checked against a list of allowed roles listed in an annotation on the pods namespace.

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -43,7 +43,7 @@ type appHandler func(http.ResponseWriter, *http.Request)
 
 // ServeHTTP implements the net/http server Handler interface.
 func (fn appHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	log.Infof("Requesting %s", r.RequestURI)
+	log.Debugf("Requesting %s", r.RequestURI)
 	log.Debugf("RemoteAddr %s", parseRemoteAddr(r.RemoteAddr))
 	w.Header().Set("Server", "EC2ws")
 	fn(w, r)
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -81,6 +81,23 @@ func (s *Server) getRole(IP string) (string, error) {
 	return role, nil
 }
 
+func (s *Server) debugStoreHandler(w http.ResponseWriter, r *http.Request) {
+	output := make(map[string]interface{})
+
+	output["rolesByIP"] = s.store.DumpRolesByIP()
+	output["rolesByNamespace"] = s.store.DumpRolesByNamespace()
+	output["namespaceByIP"] = s.store.DumpNamespaceByIP()
+
+	o, err := json.Marshal(output)
+	if err != nil {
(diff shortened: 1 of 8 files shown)


whizard:
That works well if there are multiple roles for a namespace. The other way (my PR) is a bit "magical"/hackish.

preflightsiren:
So what happens to a pod with an annotation that doesn't correspond to the annotation in the namespace?

whizard:
It can't assume a role.  The idea is to control what rights a pod can
assume in AWS and that is best (IMHO) handled/controlled at the namespace
level.

Paul

SleepyBrett:
@preflightsiren 
https://github.com/jtblin/kube2iam/pull/39/commits/11b4195633bfbfd476e158094e89597dfaa94134#diff-1427b112aa72f64adcf839007e954d56R110

If namespace restrictions are enabled and if you try to assume a role you don't have perms to.

SleepyBrett:
@jtblin, when you get a chance.

whizard:
@SleepyBrett I like the configmap approach. However, I am not sure what you mean by having multiple namespaces ("specialnamespace, namespacetwo...") since the configmap would only apply to one namespace. Am I missing something?

SleepyBrett:
@whizard One centralized configmap.

Here at my company we have one big AWS account that everyone uses. We have a group responsible for managing it and creating some extra permissions enforcement. If We had one configmap per namespace IN that namespace people in the namespaces could 'claim' roles from other teams.

It leaves it up to us, as kube admins, to confirm the roles a namespace would like to use are derived from their team role (we could probably automate this too, but I'm not sure that's really pull requestable) and then add it to the configmap (or namespace in the current PR).

SleepyBrett:
@jtblin if these changeset is a little to big to digest, let me know. I'm willing to do a skype/hangouts type code review if you want.

fiunchinho:
Any news on this? We have a multi-tenant cluster and we are really interested on this feature to be merged

jtblin:
Very sorry for the lag, been really busy lately. Will look into this next week. Do you mind resolving the conflict in the mean time?

Thanks for your patience.

cmd/server.go
```diff
@@ -79,6 +81,23 @@ func (s *Server) getRole(IP string) (string, error) {
 	return role, nil
 }
 
+func (s *Server) debugStoreHandler(w http.ResponseWriter, r *http.Request) {
```
negz: <3 I was just thinking about writing the same functionality while attempting to debug the store this weekend.

struz:
I'm interested in this functionality as well, and I made some extensions to this PR to fix the conflicts and make sure it plays nicely with the ARN changes from https://github.com/jtblin/kube2iam/pull/53.

The PR is at https://github.com/jtblin/kube2iam/pull/56

PR #42, opened 2016-12-29
added host interface name for weave

OwainPerry:


jtblin:
Thanks for sending the PR! 👍

PR #43, opened 2016-12-30
Truncating session names to avoid AWS errors

robinpercy:
Session names have a poorly-documented 64 char limit, which can cause
role assumption to fail when long role names are used. Here's the error I was seeing:

> level=error msg="Error assuming role ValidationError: 1 validation error detected: Value 'route53_kubernetes.k8s.robin.region1.playground.mydomain.org-b4952ff8' at 'roleSessionName' failed to satisfy constraint: Member must have length less than or equal to 64\n\tstatus code: 400, request id: 1f22920b-cdcf-11e6-a283-abcdefg48078f"

I've also moved the hash to the front of the session name because it
simplifies chopping off the end of the role name when necessary. That's
based on a very biased assumption that role name prefixes are more
informative than suffixes.

### DIFF
--- a/cmd/iam.go
+++ b/cmd/iam.go
@@ -15,7 +15,8 @@ import (
 var cache = ccache.New(ccache.Configure())
 
 const (
-	ttl = time.Minute * 15
+	ttl               = time.Minute * 15
+	maxSessNameLength = 64
 )
 
 type iam struct {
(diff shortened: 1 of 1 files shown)


robinpercy:
@jtblin: any thoughts?

jtblin:
Looks good, sorry for the lag but I was in annual leave. Thanks for the PR! 👍

robinpercy:
No worries. Cheers!

mikeyhu:
Are you planning on releasing a new version with this fix in soonish? CloudFormation tends to give roles created quite long names so I've run into this issue using the latest (0.3.0) image. I can work around the issue for now but it would be great to get an updated image!

pawelprazak:
is there any ETA on a the release? 

we have long dynamic role names, so we would be grateful for a release

jtblin:
Sorry guys, been pretty busy on other things lately. I've released new versions with this change now.

PR #51, opened 2017-02-27
Allow roles to be annotated with with the base name or the full ARN

negz:
I spent a long time trying to figure out why my role was not being granted despite having annotated my pods. It turns out I had annotated them with the full ARN. I've expanded the logging a little to help with similar issues in future (typos, etc) and refactored the role check slightly to allow cases where a pod is annotated with the full ARN rather than the role name alone.

Note I have not tested my changes aside from ensuring the code builds.


jtblin:
Thanks @negz ! I've merged #53 which includes your changes so will decline this one.

negz:
Thanks @jtblin and @struz!

PR #53, opened 2017-03-06
Allow full ARN specification for roles, including cross AWS account

struz:
I built on top of what @negz made (https://github.com/jtblin/kube2iam/pull/51) to fix some bugs and test that roles could be assumed between two AWS accounts assuming the trust roles were set up properly.

My main driver behind this PR was to allow the assumption of a role from another AWS account, which can only be done with a full ARN specifier.

I tested a bunch of combinations including:

- no base ARN + full ARN default role + no annotation
- no base ARN + full ARN default role + full ARN annotation
- base ARN + full ARN default role + no annotation
- base ARN + full ARN default role + full ARN annotation
- base ARN + full ARN default role + simple role name annotation

Seems to be working fine.

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -92,6 +92,7 @@ func (s *Server) securityCredentialsHandler(w http.ResponseWriter, r *http.Reque
 	if s.iam.baseARN == "" && strings.HasPrefix(roleARN, s.iam.baseARN) {
 		idx := strings.LastIndex(roleARN, "/")
 		write(w, roleARN[idx+1:])
+		return
 	}
 	write(w, roleARN)
 }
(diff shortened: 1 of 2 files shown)


cmd/server.go
```diff
@@ -151,7 +158,7 @@ func (s *Server) Run(host, token string, insecure bool) error {
 	s.iam = newIAM(s.BaseRoleARN)
 	r := mux.NewRouter()
 	r.Handle("/{version}/meta-data/iam/security-credentials/", appHandler(s.securityCredentialsHandler))
-	r.Handle("/{version}/meta-data/iam/security-credentials/{role}", appHandler(s.roleHandler))
+	r.Handle("/{version}/meta-data/iam/security-credentials/{role:.*}", appHandler(s.roleHandler))
```
jtblin: This doesn't seem to change anything i.e. `.*` will catch everything anyway.

struz: Default behaviour is to catch until the next slash (e.g. {role}), this needs to catch an ARN which may or may not have slashes in it hence the regex.

PR #56, opened 2017-03-08
Namespace restriction: revised

struz:
Update of https://github.com/jtblin/kube2iam/pull/39 to get it working. Only minor tweaks based on conflicts and merges that happened since @SleepyBrett last committed an update.

Apologies about the merge spam commits - hindsight says I should have rebased.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,7 @@ different than `docker0` depending on which virtual network you use e.g.
 * for Calico, use `cali+` (the interface name is something like cali1234567890
 * for kops (on kubenet), use `cbr0`
 * for CNI, use `cni0`
+* for weave use `weave`
 
 ```
 ---
@@ -241,4 +242,4 @@ Jerome Touffe-Blin, [@jtblin](https://twitter.com/jtblin), [About me](http://abo
 # License
 
 kube2iam is copyright 2016 Jerome Touffe-Blin and contributors.
-It is licensed under the BSD license. See the include LICENSE file for details.
+It is licensed under the BSD license. See the included LICENSE file for details.
--- a/README.md
+++ b/README.md
@@ -190,6 +190,23 @@ spec:
 
 You can use `--default-role` to set a fallback role to use when annotation is not set.
 
+### Namespace Restrictions
+
+By using the flag --namespace-restrictions you can enable a mode in which the roles that pods can assume is restricted by an annotation on the pod's namespace. This annotation should be in the form of a json array.
+
+To allow the aws-cli pod specified above to run in the default namespace your namespace would look like the following.
+
+```
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  annotations:
+    kube2iam/allowed-roles: |
+      ["role-name"]
+  name: default
+```
+
 ### Options
 
 By default, `kube2iam` will use the in-cluster method to connect to the kubernetes master, and use the `iam.amazonaws.com/role`
@@ -210,6 +227,8 @@ Usage of kube2iam:
       --insecure                Kubernetes server should be accessed without verifying the TLS. Testing only
       --iptables                Add iptables rule (also requires --host-ip)
       --metadata-addr string    Address for the ec2 metadata (default "169.254.169.254")
+      --namespace-key string    Namespace annotation key used to retrieve the IAM roles allowed (value in annotation should be json array) (default "kube2iam/allowed-roles")
+      --namespace-restrictions  Enable namespace restrictions
       --verbose                 Verbose
       --version                 Print the version and exits
 
--- a/README.md
+++ b/README.md
@@ -203,7 +203,7 @@ apiVersion: v1
 kind: Namespace
 metadata:
   annotations:
-    kube2iam/allowed-roles: |
+    iam.amazonaws.com/allowed-roles: |
       ["role-name"]
   name: default
 ```
@@ -235,7 +235,7 @@ Usage of kube2iam:
       --insecure                Kubernetes server should be accessed without verifying the TLS. Testing only
       --iptables                Add iptables rule (also requires --host-ip)
       --metadata-addr string    Address for the ec2 metadata (default "169.254.169.254")
-      --namespace-key string    Namespace annotation key used to retrieve the IAM roles allowed (value in annotation should be json array) (default "kube2iam/allowed-roles")
+      --namespace-key string    Namespace annotation key used to retrieve the IAM roles allowed (value in annotation should be json array) (default "iam.amazonaws.com/allowed-roles")
       --namespace-restrictions  Enable namespace restrictions
       --verbose                 Verbose
       --version                 Print the version and exits
--- a/README.md
+++ b/README.md
@@ -118,11 +118,11 @@ iptables \
   --to-destination `curl 169.254.169.254/latest/meta-data/local-ipv4`:8181
 ```
 
-This rule can be added automatically by setting `--iptables=true`, setting the `HOST_IP` environment 
+This rule can be added automatically by setting `--iptables=true`, setting the `HOST_IP` environment
 variable, and running the container in a privileged security context.
 
-Note that the interface `--in-interface` above or using the `--host-interface` cli flag may be 
-different than `docker0` depending on which virtual network you use e.g. 
+Note that the interface `--in-interface` above or using the `--host-interface` cli flag may be
+different than `docker0` depending on which virtual network you use e.g.
 
 * for Calico, use `cali+` (the interface name is something like cali1234567890
 * for kops (on kubenet), use `cbr0`
@@ -208,6 +208,12 @@ metadata:
   name: default
 ```
 
+### Debug
+
+By using the --debug flag you can enable some extra features making debugging easier:
+
+- `/debug/store` endpoint enabled to dump knowledge of namespaces and role association.
+
 ### Options
 
 By default, `kube2iam` will use the in-cluster method to connect to the kubernetes master, and use the `iam.amazonaws.com/role`
@@ -221,6 +227,7 @@ Usage of kube2iam:
       --api-token string        Token to authenticate with the api server
       --app-port string         Http port (default "8181")
       --base-role-arn string    Base role ARN
+      --debug                   Enable some debug features
       --default-role string     Fallback role to use when annotation is not set
       --host-interface string   Host interface for proxying AWS metadata (default "docker0")
       --host-ip string          IP address of host
--- a/cmd/k8s.go
+++ b/cmd/k8s.go
@@ -40,6 +40,26 @@ func (k8s *k8s) watchForPods(podManager kcache.ResourceEventHandler) kcache.Stor
 	return podStore
 }
 
+// returns a listwatcher of namespaces
+func (k8s *k8s) createNamespaceLW() *kcache.ListWatch {
+	return kcache.NewListWatchFromClient(k8s, "namespaces", api.NamespaceAll, selector.Everything())
+}
+
+func (k8s *k8s) watchForNamespaces(nsManager kcache.ResourceEventHandler) kcache.Store {
+	nsStore, nsController := kcache.NewInformer(
+		k8s.createNamespaceLW(),
+		&api.Namespace{},
+		resyncPeriod,
+		kcache.ResourceEventHandlerFuncs{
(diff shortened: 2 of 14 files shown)


cmd/k8s.go
```diff
@@ -40,6 +40,26 @@ func (k8s *k8s) watchForPods(podManager kcache.ResourceEventHandler) kcache.Stor
...
+func (k8s *k8s) watchForNamespaces(nsManager kcache.ResourceEventHandler) kcache.Store {
+	nsStore, nsController := kcache.NewInformer(
+		k8s.createNamespaceLW(),
+		&api.Namespace{},
+		resyncPeriod,
+		kcache.ResourceEventHandlerFuncs{
+			AddFunc:    nsManager.OnAdd,
+			DeleteFunc: nsManager.OnDelete,
+			UpdateFunc: nsManager.OnUpdate,
+		},
```
ash2k: Just pass `nsManager` directly, no need to wrap it in another layer of indirection.

cmd/namespace.go
```diff
@@ -0,0 +1,83 @@
...
+	ns, ok := obj.(*api.Namespace)
+	if !ok {
+		log.Errorf("Expected Namespace but OnAdd handler received %+v", obj)
+		return
+	}
+
+	log.Debugf("Namespace OnAdd %s", ns.GetName())
+
+	roles := h.getRoleAnnotation(ns)
+	if roles != nil {
```
ash2k: This check is unnecessary.

cmd/namespace.go
```diff
@@ -0,0 +1,83 @@
...
+		for i := range roles {
```
ash2k: `for _, role := ....` and use `role` in the loop body

cmd/namespace.go
```diff
@@ -0,0 +1,83 @@
...
+	"k8s.io/kubernetes/pkg/api"
+)
+
+type namespacehandler struct {
+	storage *store
+}
+
+// OnAdd called with a namespace is added to k8s
+func (h *namespacehandler) OnAdd(obj interface{}) {
+	ns, ok := obj.(*api.Namespace)
```
ash2k: I think it is ok to just cast "dangerously" - if some other type of object has been passed that is a serious bug somewhere and program should crash.

cmd/server.go
```diff
@@ -79,6 +82,23 @@ func (s *Server) getRole(IP string) (string, error) {
...
+func (s *Server) debugStoreHandler(w http.ResponseWriter, r *http.Request) {
+	output := make(map[string]interface{})
+
+	output["rolesByIP"] = s.store.DumpRolesByIP()
+	output["rolesByNamespace"] = s.store.DumpRolesByNamespace()
+	output["namespaceByIP"] = s.store.DumpNamespaceByIP()
+
+	o, err := json.Marshal(output)
+	if err != nil {
+		log.Errorf("Error converting debug map to json: %+v", err)
```
ash2k: This block should also return HTTP 500 instead of falling through and writing uninitialized `o` and returning 200 then.

cmd/server.go
```diff
@@ -79,6 +82,23 @@ func (s *Server) getRole(IP string) (string, error) {
...
+	}
+
+	if _, err := w.Write(o); err != nil {
+		log.Errorf("Error writing response: %+v", err)
```
ash2k: This is unnecessary - most likely client has just disconnected and there is no point logging that. Not on `error` in any case.

cmd/server.go
```diff
@@ -99,17 +119,24 @@ func (s *Server) securityCredentialsHandler(w http.ResponseWriter, r *http.Reque
...
-	allowedRoleARN := s.iam.roleARN(allowedRole)
+	podRole, err := s.getRole(remoteIP)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusNotFound)
 		return
 	}
+	podRoleARN := s.iam.roleARN(podRole)
+
+	if !s.store.CheckNamespaceRestriction(podRoleARN, remoteIP) {
+		http.Error(w, fmt.Sprintf("Role requested %s not valid for namespace of pod at %s", podRole, remoteIP), http.StatusNotFound)
```
ash2k: Maybe put pod name instead of ip? and also mention namespace

cmd/server.go
```diff
@@ -119,7 +146,7 @@ func (s *Server) roleHandler(w http.ResponseWriter, r *http.Request) {
 
 	credentials, err := s.iam.assumeRole(wantedRoleARN, remoteIP)
 	if err != nil {
-		log.Errorf("Error assuming role %+v", err)
+		log.Errorf("Error assuming role %+v for pod at %s", err, remoteIP)
```
ash2k: Pod name+namespace?

cmd/store.go
```diff
@@ -31,64 +35,132 @@ func (s *store) Get(IP string) (string, error) {
...
+			c = false
+			break
 		}
 	}
+	if c {
+		ar = append(ar, roleARN)
+	}
+	s.mutex.Lock()
+	s.rolesByNamespace[namespace] = ar
+	s.mutex.Unlock()
```
ash2k: Why all mutex unlocks in this file are not using `defer`?

cmd/store.go
```diff
@@ -31,64 +35,132 @@ func (s *store) Get(IP string) (string, error) {
...
-		defaultRole: defaultRole,
-		iamRoleKey:  key,
-		rolesByIP:   make(map[string]string),
+		defaultRole:          defaultRole,
+		iamRoleKey:           key,
+		namespaceKey:         namespaceKey,
+		namespaceRestriction: namespaceRestriction,
+		rolesByIP:            make(map[string]string),
+		rolesByNamespace:     make(map[string][]string),
+		namespaceByIP:        make(map[string]string),
```
ash2k: I think the whole store should not exist. It should be implementable via IndexedInformers with custom indexes for these ^ things. Then you don't need to maintain these mappings yourself. However this may be done in a separate PR later if this works for you.

cmd/namespace.go
ash2k: Same in other places.

cmd/namespace.go
ash2k: Same in other places.

cmd/namespace.go
ash2k: Same in other places.

cmd/namespace.go
jtblin: I'd keep the cast check here for consistency with other places. We can change that later everywhere if we want to. Same as kubernetes code base btw.

cmd/store.go
jtblin: Yeah there's another PR that was proposing that. But it would need serious rebasing and does not include namespaces obviously. Not sure why they don't use that in the kubernetes code base for DNS though and if there is any trade-off.

ash2k: Yeah, I saw that pr. It is a good idea. My guess would be because informers were not available back then. Informers are the right way to write controllers in most cases. See for more info https://github.com/kubernetes/community/blob/master/contributors/devel/controllers.md

jtblin: We already use informers (which the doc you link to seems to be about), the other PR and I assume your comment about the store are for IndexerInformer ;)

Anyhow, agreed. Let's finish this guy then we can move both the standard path and namespace to IndexerInformer.

cmd/server.go
struz: We don't actually store/track the pod name anywhere. Perhaps we could do this in another PR.

cmd/server.go
struz: See other comment regarding pod name. I will include the namespace though.

cmd/k8s.go
struz: I'm not really sure what you mean here. All the examples I can find specify functions specifically, but I am probably missing something.

cmd/namespace.go
struz: I'll keep it for now.

cmd/store.go
struz: I think it's really just a style thing. As far as I can tell for simple things like this, explicit lock/unlock code will be ever so slightly more performant but in reality it's probably negligible. Happy to change this if people think it looks better, otherwise I'll stick with what was already there.

README.md
```diff
@@ -191,6 +191,29 @@ spec:
...
+
+To allow the aws-cli pod specified above to run in the default namespace your namespace would look like the following.
+
+```
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  annotations:
+    kube2iam/allowed-roles: |
```
jtblin: Shouldn't we use something more consistent with the non namespace option e.g. `iam.amazonaws.com/allowed-roles`

cmd/namespace.go
```diff
@@ -0,0 +1,83 @@
...
+		return
+	}
+	log.Debugf("Namespace OnDelete %s", ns.GetName())
+	h.storage.DeleteNamespace(ns.GetName())
+}
+
+// getRoleAnnotations reads the "kube2iam/roles" annotation off a namespace
+// and splits them as a JSON list (["role1", "role2", "role3"])
+func (h *namespacehandler) getRoleAnnotation(ns *api.Namespace) []string {
+	rolesstring := ns.Annotations[h.storage.namespaceKey]
```
jtblin: Use camel case for "rolesstring" or just "roles".

struz:
Okay I've updated the branch addressing the requested changes now.

cmd/k8s.go
ash2k:
```
nsStore, nsController := kcache.NewInformer(
		k8s.createNamespaceLW(),
		&api.Namespace{},
		resyncPeriod,
		nsManager
)
```

ash2k: `NewInformer()` wants `ResourceEventHandler` as its last argument and `nsManager` already implements that interface so you can just pass it directly.

cmd/server.go
```diff
@@ -79,6 +82,22 @@ func (s *Server) getRole(IP string) (string, error) {
...
+	output := make(map[string]interface{})
+
+	output["rolesByIP"] = s.store.DumpRolesByIP()
+	output["rolesByNamespace"] = s.store.DumpRolesByNamespace()
+	output["namespaceByIP"] = s.store.DumpNamespaceByIP()
+
+	o, err := json.Marshal(output)
+	if err != nil {
+		log.Errorf("Error converting debug map to json: %+v", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
ash2k: Well, you need to exit the function after this line :)

struz:
And again. Let me know if there's anything major outstanding.

cmd/store.go
```diff
@@ -31,64 +35,133 @@ func (s *store) Get(IP string) (string, error) {
...
+
+// AddRoleToNamespace takes a role name and adds it to our internal state
+func (s *store) AddRoleToNamespace(namespace string, role string) {
+	// Make sure to add the full ARN of roles to ensure string matching works
+	roleARN := s.iam.roleARN(role)
+
+	ar := s.rolesByNamespace[namespace]
+	if ar == nil {
+		ar = []string{}
 	}
```
ash2k: This block is unnecessary - appending to an empty slice is ok - it will allocate memory/etc automatically.

cmd/store.go
```diff
@@ -31,64 +35,133 @@ func (s *store) Get(IP string) (string, error) {
...
-		if dok {
-			pod, ok = deletedObj.Obj.(*api.Pod)
+// checkRoleForNamespace checks the 'database' for a role allowed in a namespace,
+// returns true if the role is found, otheriwse false
+func (s *store) checkRoleForNamespace(role string, namespace string) bool {
+	ar := s.rolesByNamespace[namespace]
+	if ar == nil {
+		log.Warnf("Role:%s on namespace:%s not found.", role, namespace)
+		return false
+	}
```
ash2k: This `if` block is unnecessary, can be just deleted.

cmd/store.go
```diff
@@ -31,64 +35,133 @@ func (s *store) Get(IP string) (string, error) {
...
+	for i := range ar {
```
ash2k: `for _, r := range ar` then `if r == role`.

ash2k:
I left a few more comments, but no issues so I approved the PR 👍

PR #59, opened 2017-03-15
Adding regexes wildcard capabilities to namespace restriction

Fsero:
If your IAM Roles comes from Cloudformation predicts the specific name could be problematic, these PR enables the cluster admin to allow IAM roles that adheres to some convention, I.E if you have the convention that team name should be part of the cloudformation stack name instead of listing

teamA-powerful-role-LR34563D
you will list
teamA-powerful-role-*
or teamA-*

Of course nothing prevents to specify * which is the behaviour of kube2iam without namespace restrictions.

### DIFF
--- a/cmd/store.go
+++ b/cmd/store.go
@@ -6,6 +6,7 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 	"k8s.io/kubernetes/pkg/api"
+	"regexp"
 )
 
 // store implements the k8s framework ResourceEventHandler interface.
@@ -109,6 +110,12 @@ func (s *store) DeleteNamespace(namespace string) {
 func (s *store) checkRoleForNamespace(role string, namespace string) bool {
 	ar := s.rolesByNamespace[namespace]
 	for _, r := range ar {
+		re := regexp.MustCompile(r)
+		if re.MatchString(role) {
+			log.Debugf("Role:%s matching regexp %s on namespace:%s found.", role, r, namespace)
+			return true
+		}
+
 		if r == role {
 			log.Debugf("Role:%s on namespace:%s found.", role, namespace)
 			return true
--- a/cmd/store.go
+++ b/cmd/store.go
@@ -2,24 +2,24 @@ package cmd
 
 import (
 	"fmt"
+	"regexp"
 	"sync"
 
 	log "github.com/Sirupsen/logrus"
-	"k8s.io/kubernetes/pkg/api"
-	"regexp"
+	"k8s.io/client-go/pkg/api/v1"
 )
 
 // store implements the k8s framework ResourceEventHandler interface.
 type store struct {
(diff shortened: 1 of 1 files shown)


cmd/store.go
```diff
@@ -109,6 +110,12 @@ func (s *store) DeleteNamespace(namespace string) {
 func (s *store) checkRoleForNamespace(role string, namespace string) bool {
 	ar := s.rolesByNamespace[namespace]
 	for _, r := range ar {
+		re := regexp.MustCompile(r)
```
struz: Instead of compiling the regex every time we check for a role here, make it a member of the store struct and compile it on initialization of the store, or when the namespace is registered.

cmd/store.go
```diff
@@ -109,6 +110,12 @@ func (s *store) DeleteNamespace(namespace string) {
...
+		if re.MatchString(role) {
+			log.Debugf("Role:%s matching regexp %s on namespace:%s found.", role, r, namespace)
+			return true
+		}
+
 		if r == role {
```
struz: We can do away with this if statement here because a regex that is exactly the same as a role will match that role. As an aside, you need to change the README to ensure users know that the namespace role restriction annotation is interpreted as a regex since some characters in some ARNs may need escaping (e.g. *).

Fsero: Thanks for the feedback @struz i will implement the changes and let you know :)

jtblin:
That could be a breaking change.

On Thu, Mar 16, 2017 at 10:48 AM, Matt Whittington <notifications@github.com

Fsero:
@struz ive updated the PR with the changes requested, could you review it please? :)

Fsero:
@struz @jtblin I have this working on my cluster but I would appreciate your input here for merging this :bow:

struz:
@jtblin can you elaborate on your concerns about the change? Is it to do with race conditions around storing the regex?

jtblin:
My concern is about escaping characters e.g. / that could be specified today by providing the full arn -- or is that only possible for pod roles? It may be a good idea to write test to cover a few test cases as well.

Fsero:
I'm happy to include some tests if that makes the change more likely to be accepted :)

mumoshu:
@struz Did you have any chance to review/adapt this PR, or anything I can do to make it happen? 😃

jtblin:
@mumoshu I think the main problem with this approach is that it wouldn't be backward compatible e.g. `/` needs to be escaped for a regex so people using that today would not be able to upgrade without changing their config. It was also lacking tests. Anyhow given the heavy refactoring happening right now for using index informers, the entire store package is going to disappear so this PR is probably outdated now.

We should discuss and agree the approach on #88 before continuing.

Fsero:
Thanks for the update @jtblin :) i will put my input on #88

PR #62, opened 2017-04-06
Adding in annotation awareness to the OnChange listener

### LINKED ISSUE #61: Support adding/modifying annotations for existing pods
First off, thanks for the project, we have found it useful and to the point!

**Issue:**
When you annotate an existing pod, Kube2IAM will continue to serve up the cached role/credentials for the pod.  

In our case, we are attempting to migrate an existing stack over to Kube2IAM with as little impact as possible.  This has meant annotating some pods that are part of a stateful set via the `kubectl annotate` command. 

**Request:**
It would be very useful if the addition of or modification of a "relevant" annotation triggered a flush of the kube2iam cache

jrnt30:
Fixes #61 - When a pod changes, examines the relevant role annotation on the pod.
If it has been added, removed or updated then assumes role needs to be reassigned

### DIFF
--- a/cmd/pod.go
+++ b/cmd/pod.go
@@ -45,7 +45,7 @@ func (p *podHandler) OnUpdate(oldObj, newObj interface{}) {
 		return
 	}
 
-	if keysDiffer(oldPod.Annotations, newPod.Annotations, []string{p.storage.iamRoleKey}) {
+	if annotationDiffers(oldPod.Annotations, newPod.Annotations, p.storage.iamRoleKey) {
 		log.Debugf("Updating pod %s due to added/updated annotation value", newPod.GetName())
 		p.OnDelete(oldPod)
 		p.OnAdd(newPod)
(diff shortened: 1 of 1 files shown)


struz:
This looks good to me. Thoughts @jtblin?

cmd/pod.go
```diff
@@ -67,6 +75,17 @@ func (p *podHandler) OnDelete(obj interface{}) {
 	}
 }
 
+func keysDiffer(oldAnnotations, newAnnotations map[string]string, keysThatMatter []string) bool {
```
jtblin: 1. Why using an array of string as we'll always pass only one key?
1. It's not the key that differs IIUC, it's the annotation value so I would rename the function to `annotationDiffers`

jrnt30: Just trying to make it usable if there were additional annotations that were used in the future, not specific reason.  I'll rename and change.

jrnt30:
@jtblin Those should be addressed now.

jtblin:
Thanks a ton!

jrnt30:
Thanks for the very useful project and review.

Issue #13, opened 2016-09-16
kube2iam should discover AWS account ID


tazjin:
If no `base-role` is set, kube2iam should probably discover the AWS account ID and use `arn:aws:iam::${AWS_ACCOUNT_ID}:role/` as the default.

jtblin:
I don't think this should be the default behaviour but maybe if a `--discover-role` flag is set, we could do that. I am not familiar with this default role, do you have some doc that you can point to?

ajohnstone:
Can use the api call on sts to fetch this.
```
aws sts get-caller-identity
{
    "Account": "<ACCOUNT_ID>", 
    "UserId": "<USER_ID>", 
    "Arn": "arn:aws:iam::<ACCOUNT_ID>:user/<USERNAME>"
}

```

Would be awesome to add this.

tazjin:
@jtblin That's just the default base ARN at which a role will be located if it's not explicitly put into any path.

jtblin:
I don't think we'd want to set the default role automatically from a security standpoint, maybe we could do that with (yet) another flag e.g. `--discover-default-role` or something like that.

jtblin:
Now released as `0.5.0`. Thanks @jescarri for the contribution!

PR #63, opened 2017-04-06
Add default IAM role auto discovery

jescarri:
Related to:
https://github.com/jtblin/kube2iam/issues/13

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,6 +214,14 @@ By using the --debug flag you can enable some extra features making debugging ea
 
 - `/debug/store` endpoint enabled to dump knowledge of namespaces and role association.
 
+### Base ARN auto discovery
+
+By using the --auto-discover-base-arn flag, kube2iam will auto discover the base arn via the ec2 metadata service.
+
+### Using ec2 instance role as default role
+
+By using the --auto-discover-default-role flag, kube2iam will auto discover the base arn and the iam role attached to the instance and use it as the fallback role to use when annotation is not set.
+
 ### Options
 
 By default, `kube2iam` will use the in-cluster method to connect to the kubernetes master, and use the `iam.amazonaws.com/role`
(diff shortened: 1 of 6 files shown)


jescarri:
Hi any update on this?

jtblin:
@jescarri if you just assign the default IAM role to containers that  are running on the worker, what is the point of running kube2iam which prevents using the default role and add this option to use the default role?

jescarri:
@jtblin mi use case is as follows:
1. I have a dozen of clusters all spread across multiple accounts so base ARN changes
2. I have mixed workloads one of those workloads is prometheus + ec2 discovery
3. EC2 default role allows assumeRole and ec2:describe-instances this is used mainly for prometheus, our prometheus users use prometheus as a service and IAM is abstracted from them, this is the main reason for auto discovery the instance iam role, that way our prometheus users can just deploy prometheus to any cluster and not worry about the correct iam role for the cluster.

We use Kube2Iam as intended, pods that require more than ec2:describe-instances, need to go to the process of annotating their pods, establishing trust relationships etc, this PR is just for convenience of deployment and basic usage.

What I think is missing is the ability to auto detect the base arn and being able to specify the default iam role.

Does this makes sense?, I have been running the code of this PR in prod a couple of weeks, w/o issues.

jtblin:
Thanks for the explanation, makes sense. Didn't think about this use case.

cmd/iam.go
```diff
@@ -51,6 +54,40 @@ func getHash(text string) string {
...
+func GetInstanceIamRole() (string, string, error) {
+	sess, err := session.NewSession()
+	if err != nil {
+		return "", "", err
+	}
+	metadata := ec2metadata.New(sess)
+	if !metadata.Available() {
+		return "", "", fmt.Errorf("EC2 Metadata is not available, are you running on EC2?")
+	}
+	resp, err := http.Get("http://169.254.169.254/latest/meta-data/iam/security-credentials/")
```
jescarri: Let me fix the http request thing, I wanted to be consistent with the current operation mode, same as the --default-role but auto discovered and set to the iam instance role.

The base arn is crucial, because it changes from account to account, we could have standard IAM role names but the base will change, keeping track of those base arn is a pain.

What I propose is to break this into two flags:
--auto-discover-base-arn
--auto-discover-default-role

jtblin:
Makes sense, I forgot the base arn is mandatory, and this will fix #13.
Maybe break that into 2 PRs then, it will be easier to review, starting by
the one for the base arn?

jescarri:
sure, will have both of them in the next day or so.

README.md
```diff
@@ -1,3 +1,4 @@
+
```
jescarri: Fixed

jescarri:
@jtblin I made a single PR because the  --auto-discover-default-role is dependent on  --auto-discover-base-arn feature, I also upgraded the AWS SDK version to latest and corrected the kubernetes api version in the glide.yaml file it was set to a newer version but the glide.lock file was never upgraded, using a newer k8s client library seems to break the compile.

Thanks!.

main.go
```diff
@@ -34,6 +34,35 @@ func main() {
 		version.PrintVersionAndExit()
 	}
 
+	if s.AutoDiscoverBaseArn {
+		if s.BaseRoleARN != "" {
+			log.Fatal("--auto-discover-base-arn can't be used if --base-role-arn is specified")
+		}
+		if s.DefaultIAMRole == "" {
+			log.Fatal("You need to specify the default IAM role, use --auto-discover-default-role to autodiscover it")
```
jtblin: You're setting the default role with `AutoDiscoverDefaultRole` below. Shouldn't that be reversed?

jescarri: Is just a warning, if DefaultIamRole is blank print that is a required argument and suggest the use of --auto-discover-default-role

jtblin: Well even if you set `--auto-discover-default-role` it won't work as the test is only if the default role is set which will not have happen at this stage yet. Anyhow see my other comment about why this is even mandatory (maybe I'm missing something :)?).

jtblin:
Thanks a ton for the contribution and for your patience! 👍

jescarri:
You are welcome!

whereisaaron:
I don't see it documented but `--namespace` restrictions appear to (correctly in my view) block any `--auto-discover-default-role` or `--default-role`. The default role appears to get assigned to the Pod, but then gets blocked (500) when the Pod tries to use it.

Should the default role even get assigned if it doesn't match the namespace restrictions? Maybe yes, because the namespace restrictions could change after the Pod starts?

It would be good if the README explained this under Namespace Restrictions @jtblin. I don't mind doing that but I am not clear my understanding is correct.

```
-----------------------------------------------------
DEBUG: Request ec2metadata/GetMetadata Details:
---[ REQUEST POST-SIGN ]-----------------------------
GET /latest/meta-data/iam/security-credentials/myDefaultRole HTTP/1.1
Host: 169.254.169.254
User-Agent: aws-sdk-go/1.13.34 (go1.11.5; linux; amd64)
Accept-Encoding: gzip


-----------------------------------------------------
DEBUG: Response ec2metadata/GetMetadata Details:
---[ RESPONSE ]--------------------------------------
HTTP/1.1 500 Internal Server Error
Content-Length: 96
Content-Type: text/plain; charset=utf-8
Date: Tue, 12 Mar 2019 16:56:05 GMT
Server: EC2ws
X-Content-Type-Options: nosniff
```

PR #69, opened 2017-05-04
Flannel needs cni0 not flannel0

apenney:
I'm not even sure this is universally correct, but it was under `kops` and it surprised the hell out of me.  I thought I'd add a note for the next poor soul.


jtblin:
Thanks!

PR #73, opened 2017-05-19
Replace role-name with role-arn

mikesplain:
I was a bit confused why this wasn't working at first, since I used the explicit name of the IAM role, but then realized it's probably the whole ARN.  I'm not sure if this was done for a specific reason but I figured I'd propose changing it to ARN as that's the only way I could get my roles to work.  

Thanks for the awesome project!

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ metadata:
   labels:
     name: aws-cli
   annotations:
-    iam.amazonaws.com/role: role-name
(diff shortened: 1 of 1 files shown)


jtblin:
Yeah the role arn can be retrieved automatically with the `auto-discover-base-arn` option but this change can avoid confusion. Thanks! 👍

PR #76, opened 2017-05-29
Remove runtime.GOMAXPROCS hack

dcheney-atlassian:
The runtime.GOMAXPROCS(runtime.NumCPU()) hack hasn't been needed since
Go 1.5. Let this hack rest in peace.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,8 +1,6 @@
 package main
 
 import (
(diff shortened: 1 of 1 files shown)


jtblin:
Indeed. Thanks! 👍

PR #81, opened 2017-06-08
Allow base-role-arn with added paths to be valid.

pvdvreede:
I use a path as part of the role so that we can limit access of assumption. The validation commit in 0.6.3 does not seem to allow for extra paths as I get a validation error on the base-role-arn. 

This PR extends the reg ex validation to allow arbitrary extra paths on the role.


jtblin:
Thanks for sending the PR 👍. I think this change wouldn't have allowed base arns describe in #84 neither simple ones without trailing slashes. Also I wanted to add tests so I went ahead and fix it and published 0.6.4.

PR #86, opened 2017-06-20
Fix for roles with a custom path name

totallyunknown:
It's possible to scope roles to a custom path (arn:aws:iam::123456789012:role/team-a/test-role). The current implementation doesn't support this. Instead of test-role this fix will return team-a/test-role within the securityCredentialsHandler handler. See http://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html for more details
about role paths.


jtblin:
Thanks! 👍

PR #92, opened 2017-08-12
Updating to use IndexInformer

jrnt30:
NOTE: This is the start of a rebase/rework for #32.  There ended up being some significant changes that made it difficult to integrate the changes easily and I ended up moving around a bit of the logic from store and server in an attempt to make it a bit more reusable and testable.  
Very open to feedback but wanted to get some eyes on this from @jtblin and @mikekap if they have time to ensure I'm heading down the right path.  

I have done some initial testing in our staging cluster but need to spend some more time next week going through some of the edge cases on in `server/` and test some cross-account roles tested

- Removed reliance on the store for maintain state and
migrated to the IndexInformer.  This now will only cache
"active" pods (no failed/completed pods will be returned)
- Extracted role based fallback logic to it's own processor
and removed from the Server/Event handlers
- Migrated event handlers to the k8s package
- Added in processor tests for the new RoleProcessor


jtblin:
Thanks @jrnt30, can you please revert this: 

> Moved the cmd package back to the base to allow for easy builds
and acknowledging the fact that the project itself is the unit of
deployment and it's unlikely this would be used as a library in and of
itself

I will start reviewing when this is done. It's the "standard" directory structure for go projects and I would like to keep it as-is. Moreover it creates unnecessary noise in the PR making it harder to review.

jrnt30:
@jtblin Sure thing, that is done and conflicts merged with master.

jrnt30:
Thanks for the feedback. The imports structure makes a lot of sense, those are cleaned up.  

The tests themselves were an oversight when I moved main back into the `cmd/` package but didn't change the Coveralls configuration back.  

That's been fixed as well.

processor/processor.go
```diff
@@ -0,0 +1,155 @@
...
+	"fmt"
+
+	log "github.com/sirupsen/logrus"
+	"k8s.io/client-go/pkg/api/v1"
+
+	"github.com/jtblin/kube2iam/iam"
+)
+
+// RoleProcessor handles relevant logic around associating IPs with a given IAM role
+type RoleProcessor struct {
```
jtblin: `processor.RoleProcessor` is a bit redundant and a bit of a mouthful and `processor` is too generic for a package name. It seems this is more about mapping ips and roles so something around `mapper` or `mappings` may be better suited for the package name.

processor/processor.go
```diff
@@ -0,0 +1,155 @@
...
+// checkRoleForNamespace checks the 'database' for a role allowed in a namespace,
+// returns true if the role is found, otheriwse false
+func (r *RoleProcessor) checkRoleForNamespace(roleArn string, namespace string) bool {
+	if !r.namespaceRestriction || roleArn == r.defaultRoleARN {
+		return true
+	}
+
+	ns, err := r.kubeStore.NamespaceByName(namespace)
+	if err != nil {
+		log.Debug("Unable to find an indexed namespace of [%s]", ns)
```
jtblin: `ns` is not a `string` afaict so it wouldn't print out properly and anyway it is not found. I guess you want to use `namespace` here. 

I think the [] are unnecessary and better use fields for structured logging. See *Fields convenience methods in other places.

processor/processor.go
```diff
@@ -0,0 +1,155 @@
...
+		return false
+	}
+
+	ar := GetNamespaceRoleAnnotation(ns, r.namespaceKey)
+	for _, role := range ar {
+		if r.iam.RoleARN(role) == roleArn {
+			log.Debugf("Role:%s on namespace:%s found.", roleArn, namespace)
```
jtblin: Nit: space after :

processor/processor.go
```diff
@@ -0,0 +1,155 @@
...
+			return true
+		}
+	}
+	log.Warnf("Role: [%s] on namespace: [%s] not found.", roleArn, namespace)
```
jtblin: I think the [] are unnecessary and better use fields for structured logging. See *Fields convenience methods in other places.

processor/processor_test.go
```diff
@@ -0,0 +1,241 @@
+package processor
```
jtblin: Thanks for adding all these tests! 👍 

There's no test for `GetRoleMapping`, maybe not strictly necessary though as you're testing the other methods but if not too painful could be nice.

server/server.go
```diff
@@ -279,11 +265,23 @@ func (s *Server) Run(host, token string, insecure bool) error {
...
+	podSynched := s.k8s.WatchForPods(k8s.NewPodHandler(s.IAMRoleKey))
+	namespaceSynched := s.k8s.WatchForNamespaces(k8s.NewNamespaceHandler(s.NamespaceKey))
+
+	synced := false
+	for i := 0; i < defaultCacheSyncAttempts && !synced; i++ {
+		synced = cache.WaitForCacheSync(nil, podSynched, namespaceSynched)
+	}
+
+	if !synced {
+		log.Fatalf("Attempted to wait for caches to be synced for [%d] however it is not done.  Giving up.", defaultCacheSyncAttempts)
```
jtblin: Nit `... for %d attempts, however ...`

server/server.go
```diff
@@ -279,11 +265,23 @@ func (s *Server) Run(host, token string, insecure bool) error {
...
-	s.store = model
-	s.k8s.WatchForPods(kube2iam.NewPodHandler(model))
-	s.k8s.WatchForNamespaces(kube2iam.NewNamespaceHandler(model))
+	s.roleProcessor = processor.NewRoleProcessor(s.IAMRoleKey, s.DefaultIAMRole, s.NamespaceRestriction, s.NamespaceKey, s.iam, s.k8s)
+	podSynched := s.k8s.WatchForPods(k8s.NewPodHandler(s.IAMRoleKey))
+	namespaceSynched := s.k8s.WatchForNamespaces(k8s.NewNamespaceHandler(s.NamespaceKey))
+
+	synced := false
+	for i := 0; i < defaultCacheSyncAttempts && !synced; i++ {
+		synced = cache.WaitForCacheSync(nil, podSynched, namespaceSynched)
```
jtblin: Nice, that should prevent the race condition for default role? 👍

dadux:
We've tried running of this branch in our dev environments, but still get a lot of "default role" under load (our build-engineering team spinning up 100s of concurrent jobs)

jtblin:
Thanks for the info @dadux, that's annoying. Looking at the code, it only falls back to the default role if the pod is found by its ip, there is no annotation, and there is a default role set. So if it's getting the default role when it shouldn't, it can only be because the cache has not been updated yet.

The solutions I can think of are:

1. Make a synchronous API call in case of fallback to ensure an annotation hasn't been set. There would be a latency impact obviously and a new code path to manage.
1. Move the fallback to the default role down the chain so that we return an error from `extractRoleARN`. That will trigger the exponential backoff operation retry and should be able to catch the updated annotation. Latency impact as well but at least no need for additional synchronous API call.
1. Remove the default role option as it is not reliable.

Anyhow let's **not** try to tackle that in this PR for now. We can look into it after.

cover.sh
```diff
@@ -11,13 +11,12 @@ echo "mode: count" > coverage.out
 # Initialize error tracking
 ERROR=""
 
-declare -a packages=('' \
-    'cmd' \
+declare -a packages=('cmd' \
```
jrnt30: Since there are no `*.go` files now, Coveralls will complain if the `''` is left in.  Moved the other to be sorted.

Makefile
```diff
@@ -26,10 +26,10 @@ setup:
 	gometalinter --install --update
 	glide install --strip-vendor
 
-build: *.go fmt
```
jrnt30: Since we removed the root `*.go` files it didn't trigger properly.

k8s/namespace.go
```diff
@@ -0,0 +1,82 @@
+package k8s
```
jrnt30: That's fair.

server/server.go
jrnt30: It was more around ensuring that both the pods and namespaces were indexed to ensure that you couldn't "sneak in" a role.  This only happens @ startup time, so there is the possibility that you could get the default role immediately if the new pod hasn't been indexed.  

What this PR aims to solve is a permanent caching of those incorrect roles.

processor/processor.go
```diff
@@ -0,0 +1,155 @@
...
+
+	output["rolesByIP"] = rolesByIP
+	output["namespaceByIP"] = namespacesByIP
+	output["rolesByNamespace"] = rolesByNamespace
+	return output
+}
+
+// GetNamespaceRoleAnnotation reads the "iam.amazonawr.com/allowed-roles" annotation off a namespace
+// and splits them as a JSON list (["role1", "role2", "role3"])
+func GetNamespaceRoleAnnotation(ns *v1.Namespace, namespaceKey string) []string {
```
jrnt30: After moving `namespace.go` back to the base package we can do so, yes.

processor/processor.go
jrnt30: Good catch.

jrnt30:
@jtblin Thanks for the guidance and constructive feedback, it's very helpful.  I have addressed a majority of these issues.  The `processor` renaming I have left as is for now and will come up with something better.

jrnt30:
I also know that @placydo was looking at it a bit.  Are your experiencing ok so far?

jtblin:
I've pushed docker image `jtblin/kube2iam:dev` for convenience if someone wants to try in its cluster. Would really appreciate some feedback given the size of the PR.

placydo:
Hey. Currently I am in middle of something. I will try to pull this image today and give it a try + test it over the weekend. Will come back with feedback asap. Thanks guyz!

jtblin:
@struz did you guys see any issue running this on our cluster?

placydo:
@jtblin @jrnt30 We have not seen this issue anymore. Everything works as expected!

mappings/mapper.go
```diff
@@ -0,0 +1,141 @@
...
+		if r.iam.RoleARN(role) == roleArn {
+			log.Debugf("Role: %s on namespace:%s found.", roleArn, namespace)
+			return true
+		}
+	}
+	log.Warnf("Role: %s on namespace: %s not found.", roleArn, namespace)
+	return false
+}
+
+// DumpDebugInfo outputs all the roles by IP addresr.
```
Fsero: nit: s/addresr/address/

jtblin:
Thanks a ton again @jrnt30 for the great work. This is now merged and released as `0.8.0`.

PR #95, opened 2017-09-07
Incorrect ip to namespace table for pods in a failed state.

corlettb:
We run a number of cronjobs in our environment.

These pods finish running in a Complete or Failed state and linger for sometime. Unfortunately kubernetes leaves the PodIP intact that was used for these pods.

If docker reuses these ips then we have a conflict and this leads to ips being flagged against the wrong namespace. This breaks the restricted namespace functionality.

This change restricts the namespace tracking to pods in a Running phase.

This might not be a complete solution to this issue but maybe it can start a dialogue.

### DIFF
--- a/pod.go
+++ b/pod.go
@@ -18,10 +18,11 @@ type PodHandler struct {
 
 func (p *PodHandler) podFields(pod *v1.Pod) log.Fields {
 	return log.Fields{
-		"pod.name":      pod.GetName(),
-		"pod.namespace": pod.GetNamespace(),
-		"pod.status.ip": pod.Status.PodIP,
-		"pod.iam.role":  pod.GetAnnotations()[p.storage.IamRoleKey],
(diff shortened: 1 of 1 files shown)


SleepyBrett:
I think I might be seeing this even with cronjob pods that aren't failed but completed. I'm going to evaluate this as a possible solution.

jrnt30:
@SleepyBrett That does happen.  We have a working set that changes the indexing pattern that is build with the `:dev` tag.  @jtblin has been looking for individuals to test it. 

We have been running it for several weeks now without these issues.

jtblin:
Yes #92 should fix these issues so I don't think we're going to need this PR. Going to release #92 soon now.

jtblin:
#92 is now released as `0.8.0` and should solve this issue so closing this PR.

Issue #88, opened 2017-06-26
Handling dynamic role names


lstoll:
When roles are created via CloudFormation or other dynamic tooling in an environment that can support multiple clusters per AWS account, the role names are usually either not predictable or vary between Kubernetes Clusters. This makes manifest management more complex, because the annotations are hard embedded and can't be substituted from ConfigMaps or the like. This has resulted in having to template out these fields externally, which is somewhat of a hassle.

It would be nice to have a way to use consistent arn references in manifests, and rely on kube2iam to map these to the kubernetes cluster specific values. I've had a few thoughts on how to do this.

**1. Global Prefix/Suffix**:  Allow specifying a prefix or suffix for role names as a flag to kube2iam. arn's are specified in manifests without this, and then kube2iam will apply this to the name. Pros: do not need to know all the name mappings up front. Cons: doesn't allow for mixed usage - where some roles are cluster specific, and some are not

**2. Explicit mappings as CLI flag**: Allow explicit mappings to be specific as a series of CLI flags like `--arn-alias=simple-name,arn:aws::iam:1234567890::role/stack-aasdkhkjhakjhasd-role-name`. Manifests refer to the simple name, kube2iam switches it out internall. Pros: simple, explicit, can be defined per role. Cons: need to know ahead of time the specific mappings in use, need to update the daemonset on change.

I prefer 2. It's more explicit and likely less surprising over time. These could also potentially be used to solve the issue in #59 as well. I briefly considered the idea of using a regex filter here as well, but that would make for a complex implementation and potentially surprising behaviour (e.g what happens if two roles match the regex?)

I'm happy to work up an implementation for this, I just wanted to float the idea for preliminary discussion/approval before diving in.

idubrov:
Is it possible to have an option 3, where role name is retrieved from the environment variable (which in turn will be could from the config map)?

antoniobeyah:
Our use case: We restrict teams to namespaces and only want to allow namespaces to assume roles that match their namespaces.  We are mapping namespaces to paths within IAM.  We don't really care what the actual role is as long as it is within the namespace of the team.  Role names are generated dynamically, so we don't know them ahead of time.

Our goal is to be able to annotate the namespace with an allowed roles that match the path as we define in IAM.

For example:
**Namespace:** `namespace1`
**Allowed Assumption:** `arn:aws:iam::123456789012:role/namespace1/*`
**Example IAM Roles:**
[...]

PR #104, opened 2017-10-10
Add namespace path-globbing support

### LINKED ISSUE #105: IAM roles with path components cannot be used
When using an IAM role containing a path component, the following fatal error is displayed:
`Can't determine BaseARN`.

This is because the code (incorrectly) assumes only a single `/` can exist in an IAM role ARN --see [docs](http://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html#identifiers-friendly-names).

kube2iam should be able to determine the baseARN for roles containing a path component containing one or more `/`s.

matt-deboer:
This adds the ability to use glob-based matching on namespace restrictions, which works nicely with AWS [path-based name-spacing of IAM roles](http://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html#identifiers-friendly-names).

For example, to allow pods in the default namespace to assume any role beginning with path `my/default/path/`, you would annotate the namespace like so:
```yaml
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    iam.amazonaws.com/allowed-roles: |
      ["my/default/path/*"]
  name: default
```
Potentially covers some of the discussion in #88 
Fixes #105 

**Update**: image available with this fix: `mattdeboer/kube2iam:0.6.6`


jtblin:
Sorry it took me so long to review. Looks great 👍. A negative test case for globs would have been awesome but merging as it has been so long...

PR #108, opened 2017-11-15
Limit relevant pods to those on the same node as kube2iam

mikkeloscar:
In clusters with 1000s of pods the memory used by each kube2iam
instance explodes because each kube2iam instance is storing information
about all pods on all nodes in memory.

This adds a flag `--node` for specifying the node name of kube2iam. If
set, it will limit the pods it's watching to those only on the same
node as the kube2iam instance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,12 @@ spec:
           name: kube2iam
           args:
             - "--base-role-arn=arn:aws:iam::123456789012:role/"
+            - "--node=$(NODE_NAME)"
+          env:
+            - name: NODE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
           ports:
             - containerPort: 8181
               hostPort: 8181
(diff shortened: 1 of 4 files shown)


mikkeloscar:
Build seem to be failing because of code in `vendor/`. Maybe it doesn't make sense to run lint on the `vendor/`?

jrnt30:
This seems like a solid idea to me.  I also think updating to not lint the vendor package makes sense.

hjacobs:
LGTM, let's get this merged to allow thousands of pods in a cluster :smile:

mikkeloscar:
Opened https://github.com/jtblin/kube2iam/pull/109 to fix the broken travis build.

mikkeloscar:
@jtblin done.

mikkeloscar:
and now gometalinter ran out of memory in travis \o/ @jtblin maybe it's enough to retrigger the build?

mikkeloscar:
@jtblin This is now green :)

hjacobs:
@jtblin can we get this merged?

erickt:
@jtblin With this change, would it make sense to revive https://github.com/jtblin/kube2iam/pull/30? I'm finding kube2iam occasionally taking longer than 1 second, which causes awscli to timeout.

mikkeloscar:
@jtblin Ping! :)

jtblin:
Sorry for the lag. Done.

jtblin:
Published 0.10.0 with this change.

Re: #30, I think we could look at some prefetching solution as long as it isn't going to hit AWS limits and that it is done in a memory and cpu efficient way i.e. not one goroutine per role. TBH I personally don't have the time to look into it at the moment (and in the foreseeable future).

PR #109, opened 2017-11-22
Fix govet in make check

mikkeloscar:
Changes the `go tool vet` linter command to use `govet` as introduced in https://github.com/alecthomas/gometalinter/pull/375

This makes it correctly ignore the `vendor/` folder so travis builds doesn't fail.


jtblin:
Thanks for tracking this down and fixing it! 👍

PR #116, opened 2017-12-15
Add json log formatter option

pwillie:
Will output log messages in JSON format for easy parsing by logstash or Splunk.


jtblin:
Thank you!

PR #117, opened 2017-12-19
Automatic IP tables configuration fails when using a specified metada…

imduffy15:
…ta address

Originally, the metadata address included both the host and port of the metadata
service.

The iptables rule adder had a hard coded port of 80 and would fail if the metadata
address included a port.

Added an extra paraemeter of --metadata-port to solve this.

Signed-off-by: Ian Duffy <ian.duffy@zalando.ie>


imduffy15:
Closing due to no review or follow up.

PR #128, opened 2018-02-05
Offer pod namespace and name for validation when assuming role

biochimia:
Running kube2iam in a multi-tenant setup, also connected to multiple AWS accounts, I thought it would be useful to offer the pod's namespace and name up for validation on the AWS IAM side as well. This could be extended to include some secret value in place of the `kube2iam/` prefix, but I thought I'd share early and ask for feedback on this feature.

In the default case, where no conditions on the `ExternalId` have been configured, this is harmless metadata that may be logged in CloudTrail.

### DIFF
--- a/iam/iam.go
+++ b/iam/iam.go
@@ -73,7 +73,7 @@ func sessionName(roleARN, remoteIP string) string {
 }
 
 // AssumeRole returns an IAM role Credentials using AWS STS.
-func (iam *Client) AssumeRole(roleARN, remoteIP string) (*Credentials, error) {
+func (iam *Client) AssumeRole(roleARN, remoteIP string, externalID string) (*Credentials, error) {
 	item, err := cache.Fetch(roleARN, ttl, func() (interface{}, error) {
 		sess, err := session.NewSession()
 		if err != nil {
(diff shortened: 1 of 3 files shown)


jrnt30:
Having just gone through some similar adjustments in another system for compliance purposes, I am a fan of this change conceptually however I think there is a bit of an implementation issue as it stands right now. 

Currently the cache key for the assumed role credentials do not take this attribute into account therefore we would may not have an actual 1:1 between the pods on a given node requesting that credentials and the pods that are given it.  

As an example, if I have two pods `nginx-abc123` and `nginx-dec456` that are both using the same `my-nginx-arn` and are placed on a single host the second pod that makes a request for a credential while the cache is valid (15 minutes by default) will be served by the cache and not have an external ID reflecting it's own pod name. 

The idea of better transparency between a credential and it's use for auditing I think is very valuable however the current implementation would not provide a strict enough guarantee to merge.  

A few other considerations we should take into account are the impact on the STS endpoint and API rate limiting if we do adjust the key to include the `namespace/pod name` as that will potentially increase the number of cache misses.

jrnt30:
@biochimia Any desire to provide feedback/continue this work?  If not, I will close the PR for now but feel free to open again in the future.

PR #130, opened 2018-03-14
Fix missing redirect on m5/c5 instances

mikkeloscar:
The AWS metadata service works differently on new instance types like
m5.large and c5.large.

The behavior on old instances is that if you call:

```
169.254.169.254/latest/meta-data/iam/security-credentials
```

It will redirect to:

```
169.254.169.254/latest/meta-data/iam/security-credentials/
```

Which will respond with the IAM role name.

On new instances there is no redirect, it just responds with the IAM role name directly.

Since kube2iam didn't intercept the URL without the slash, the pod will
initially get the worker node role and then try to request credentials for that
role, which kube2iam will deny.

Fix is to intercept both versions of the URL.

Fix #127

PR #131, opened 2018-03-22
Address an issue for non-AWS clusters

bigkraig:
I'm handling a 301 in kube2iam rather than using the proxy service to fetch the 301 from AWS. We are running some clusters outside of AWS but still want our pods to be able to assume IAM roles, this enables that functionality.


bigkraig:
@jtblin if you please. thanks!

jtblin:
soonSorry for the lag, I've merged #130 which does something similar and  I will deploy a new version. I don't think this one is required anymore, please reopen otherwise.

PR #146, opened 2018-06-08
Add amazon-vpc-cni-k8s to the host-interface examples

homelinen:
Adds instructions for `amazon-vpc-cni-k8s` which is the default in the recently released EKS service.

Not sure if it needs more explanation around it being in EKS, as EKS users may come here looking for help.


deiwin:
May be interesting to compare to _kiam_ documentation, which [uses `!eth0`](https://github.com/uswitch/kiam#typical-cni-interface-names).

jrnt30:
Thanks for the input, I haven't been able to validate the differences with `kiam` but *ass*uming this is working as we have other PRs for the same change as well...

PR #148, opened 2018-06-19
Update README.md to mention interface for Cilium

danwent:
The README.md currently include instructions for what interface name/regex to use when configuring kube2iam with various Kubernetes networking solutions.   This change follows the existing format and adds those instructions for http://www.cilium.io based on a question asked on the Cilium slack channel.  The user validated that this configuration works with kube2iam w/Cilium.


corlettb:
+1

jtblin:
Thanks!

PR #157, opened 2018-08-08
Regexp support for namespace restrictions (opt-in)

r4vi:
Adds support for namespace restrictions based on regexp matching instead of glob
matching. This is a backwards compatible patch which allows users to opt-in to
regex matching by specifying `--namespace-restriction-format=regexp`

This patch exists because I want to enable namespace restrictions using regex and initially I thought I would just switch to [kiam](https://github.com/uswitch/kiam), but after trying it and being put off by the complexity of it I've decided it would be easier just to implement this in kube2iam.

Added tests and make it backwards compatible so that matching defaults to glob.


r4vi:
Travis is failing because of changes in gometallinter v1 to v2, this PR: https://github.com/jtblin/kube2iam/pull/158 fixes CI

mappings/mapper.go
```diff
@@ -90,10 +93,23 @@ func (r *RoleMapper) checkRoleForNamespace(roleArn string, namespace string) boo
...
+		if strings.ToLower(r.namespaceRestrictionFormat) == "regexp" {
+			matched, err := regexp.MatchString(normalized, roleArn)
+			if err != nil {
+				log.Errorf("Namespace annotation %s caused an error when trying to match: %s for namespace: %s", rolePattern, roleArn, namespace)
+				return false
+			}
+			if matched {
+				log.Debugf("Role: %s matched %s on namespace:%s.", roleArn, rolePattern, namespace)
+			}
+			return matched
```
jrnt30: Haven't had a chance to try this out, but it seems this would "fast fail" on a set of regexps to process and the first one doesn't match.  

We should add a test case (for both Glob and Regexp) that proves out that it will attempt to match against *every* role not just the first.

r4vi: @jrnt30  yes you're right, we only ever have one annotation so I didn't think about fast fail behaviour. I'll look at improving this and adding some tests for multiple annotations

r4vi:
@jrnt30 I've made the suggested changes

mappings/mapper.go
```diff
@@ -9,16 +9,19 @@ import (
 
 	"github.com/jtblin/kube2iam"
 	"github.com/jtblin/kube2iam/iam"
+	"regexp"
```
xoen: Probably these two packages go in the first group with `fmt` as [they're standard library packages](https://github.com/golang/go/wiki/CodeReviewComments#imports).

jrnt30:
This looks good, thanks!  The import comment I do think would be "nice to have" but will look into seeing if we can automate some of those conventions as well in the future. 

Before I merge this I'd like to look at the merge/release/docker push cycle of the project to ensure I understand what is going on.  That will be a day or two unfortunately.

PR #158, opened 2018-08-08
Fix `make check` CI error

r4vi:
gometallinter doesn't install govet anymore, manually install it so that CI
doesn't fail. Once this is merged the following error should go away: https://travis-ci.org/jtblin/kube2iam/builds/413586654#L787

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@ setup:
 	go get -v -u github.com/alecthomas/gometalinter
 	go get -v -u github.com/jstemmer/go-junit-report
 	go get -v github.com/mattn/goveralls
+	go get -v github.com/dnephin/govet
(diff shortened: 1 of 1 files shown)


jrnt30:
Thanks for the PR.  I don't use the command much, is there a reason we would go with this submodule approach vs. installing the upstream command via https://github.com/golang/go/tree/master/src/cmd/vet?

r4vi:
@jrnt30 this was the most 'like-for-like' change I could think of, I didn't investigate the changes in the upstream go vet command. I do think we should use that but I'm not familiar enough with the go ecosystem to make the call. I'd suggest merging this so CI works and adding an issue/bug to move to the upstream command

PR #159, opened 2018-08-10
Avoid the following error: `ValidationError(DaemonSet.spec)`

metacyclic:
According to the [docs](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/)
> As of Kubernetes 1.8, you must specify a pod selector that matches the labels of the .spec.template. The pod selector will no longer be defaulted when left empty.

Without these changes, the following error is seen:
`error: error validating "kube2iam.yaml": error validating data: ValidationError(DaemonSet.spec): missing required field "selector" in io.k8s.api.apps.v1.DaemonSetSpec; if you choose to ignore these errors, turn validation off with --validate=false`

Adding these changes satisfies validation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,9 @@ metadata:
   labels:
     app: kube2iam
 spec:
+  selector:
+    matchLabels:
+      name: kube2iam
   template:
     metadata:
       labels:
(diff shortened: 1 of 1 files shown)


metacyclic:
Not sure why this failed to build on a `README.md` change

`The command "go get -v github.com/mattn/goveralls" failed and exited with 1 during .`

jrnt30:
Thanks for the PR!  The vet issue should be cleaned up now going forward as well.

PR #160, opened 2018-08-13
Adds prometheus metrics to kube2iam

mwhittington21:
These track:
* HTTP handler request durations and counts
* IAM request durations and counts
* Version information
* Healthcheck status
* IAM cache hit count

Metrics can either listen on the regular kube2iam port or a separate
one.


mwhittington21:
Just pushed a commit to fix the metalinter warnings which I hadn't rerun since I added the last touches and rebased.

Regarding the coverage decrease, it isn't standard procedure to test prometheus metrics code. Most of the code I added was just instrumentation and helper functions / rearchitecting to help use metrics nicely.

mwhittington21:
No problems, thanks for looking at the PR.

I just copied the histogram layout from etcd where they were mostly concerned about higher percentiles - I can change to linear if you think that would provide a better use case for HTTP distributions.

mwhittington21:
After a quick look into prometheus histograms and their applications I think this exponential approach is the most useful. This turns into the breakdown of:

```
1ms
2ms
4ms
8ms
16ms
32ms
64ms
128ms
256ms
512ms
1.024s
2.048s
4.096s
8.192s
+Inf
```

I believe that the most useful metrics there lie between 1ms and 1s, and that past that all you need to know is that your request is very slow. If we were to add more buckets it would put more strain on prometheus, and if we were to change the distribution I think it would become less useful.

jrnt30:
That looks like a reasonable distribution to start off with to me as well.  Thanks!

metrics/metrics.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// GetHandler creates a prometheus HTTP handler that will serve metrics.
+func GetHandler() http.Handler {
+	return promhttp.Handler()
+}
+
+// NewFunctionTimer creates a new timer for a generic function that can be observed to time the duration of the handler.
+// The metric is labeled with the values produced by the lvsProducer to allow for late binding of label values.
+// If provided, the timer value is stored in storeValue to allow callers access to the reported value.
```
jrnt30: Hadn't seen this model before, that is a helpful but somewhat indirect model.  Thanks for the documentation, that made it understandable.

server/server.go
```diff
@@ -309,6 +385,7 @@ func (s *Server) Run(host, token, nodeName string, insecure bool) error {
 func NewServer() *Server {
 	return &Server{
 		AppPort:                    defaultAppPort,
+        MetricsPort:                defaultAppPort
```
jrnt30: Missing `,` is prohibiting this from building for linting in Travis

mwhittington21:
Yep those were just left over from the rushed merge conflict resolution. My apologies for that. All fixed and building now!

mwhittington21:
@jrnt30 @jtblin any chance for a review this week? Looking to roll this out to our clusters without building our own image.

mwhittington21:
@jrnt30 @jtblin any chance for a review soon? Looking to roll this out to our clusters asap without building a fork if we don't have to.

jrnt30:
I will try and take another look at this later week.  

I have found in the past running my own fork for a few days has helped me find a few edge cases I wasn't expecting and our test coverage to date is a bit lacking. I would welcome some feedback on it "actually running" as well and would encourage you to build your own image and run it.

timm088:
Could use this as well, PR looks like a solid first step.

jrnt30:
@ mwhittington21 Thanks for the PR!  I got some time to read through this today and it is looking great. 

Few questions: 
- What does `lvs` stand for?  It's used a few times but unclear. I figured "label values" but not sure

A few requests and then we will be good to go: 
- Please add in the additional arguments to the README.md to outline the additional option
- Please add a section about the exported Prometheus metrics to the README (or if you think it's warranted a separate document)
- Can you document in the health handler the deferred func.  It took a few times of reading to really grok it, it's clever but could use some docs for casual reads.

mwhittington21:
@jrnt30 yeah we've been running a container built from this branch it for a little while in some of our environments. Haven't found any sharp edges yet.

`lvs` does indeed stand for label values. I just picked up the trend [from Prometheus](https://github.com/prometheus/prometheus/blob/master/vendor/github.com/prometheus/client_golang/prometheus/histogram.go#L326).

Good catch about the README, completely forgot about it. Will add that today. Same for the health handler docs.

mwhittington21:
Updated!

PR #161, opened 2018-08-14
Add Documentation & Example For EKS

asafpelegcodes:
While setting up kube2iam on my EKS cluster I struggled to get the service working until someone in the k8s slack  generously donated an hour of their time to help me debug it.
To facilitate faster setup for EKS users in the future I've added the following two changes
* Correct host interface for a default EKS cluster
* Example eks yaml file that is generic for EKS users and can be generally extended from the other documentation in the README


jrnt30:
Thanks for the PR!  I updated it to be a bit more clear that EKS is deployed with the CNI referenced but I totally see why that was confusing.

asafpelegcodes:
@jrnt30 awesome, thanks for the quick turnaround.

PR #162, opened 2018-08-20
Add new flags for session TTL

aranair:
We have long-running applications with pre-signed urls for videos. With a hard-coded 15mins TTL, videos would break, or if we did a re-sign + re-render, it would reset the video.

This PR adds a flag for the user to control the session TTL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -553,6 +553,7 @@ Usage of kube2iam:
       --backoff-max-elapsed-time duration     Max elapsed time for backoff when querying for role. (default 2s)
       --backoff-max-interval duration         Max interval for backoff when querying for role. (default 1s)
       --base-role-arn string                  Base role ARN
+      --iam-role-session-ttl                  Length of session when assuming the roles (default 15m)
       --debug                                 Enable debug features
       --default-role string                   Fallback role to use when annotation is not set
       --host-interface string                 Host interface for proxying AWS metadata (default "docker0")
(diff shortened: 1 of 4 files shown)


jrnt30:
I would like to understand this use case a little bit more before fully weighing in. As I understand it: 
- You have a pod running with a Kube2IAM annotation
- That pod will be responsible for generating pre-signed URLs which you want to have a configurable TTL on.  

Is that accurate?  If so, I don't think we would need to customize the TTL of the Pod itself.  The API call to generate the pre-signed URL itself would be where you would want to inject that TTL customization. 

The only reason I could see the two being this tightly coupled (TTL of the pod's creds vs. the TTL of the signed URL) would be if AWS expired pre-signed URLs or limited the TTL of the pre-signed URL to that of the token.  I don't *think* AWS makes that sort of limitation do they? That would complicate a lot of workloads.  If we relate this back to "standard" EC2 metdata refreshing their lease behind the scenes on their credentials, it makes me fairly confident the two are not tied together. 

It's distinctly possible I am missing something and would be happy to discuss/understand better.

aranair:
From my testing, it seems like they do even if the expiration is later than the token. I've been signing with an expiration in year 2039 and urls signed with temporary credentials (e.g. w the session tokens, as recommended) will display something along the lines of - `The provided token has expired` - instead of `Request has expired` when its the expiration of the url itself.

https://github.com/aws/aws-sdk-ruby/issues/808 and related threads in there seem to suggest this too.

aranair:
Ah, found an official response here: https://forums.aws.amazon.com/thread.jspa?threadID=153786

jrnt30:
Thanks, very interesting. I'll take a look at the full PR.

jrnt30:
Merged in a few other PRs today and forgot about this one.  Sorry.  I will cut a new release with this next week.

aranair:
np, thanks @jrnt30! Just resolved the conflicts as well -

jrnt30:
Thank you for your first contribution and fixing those conflicts.  It's merged now and I will cut a new 0.10.4 release with this next week.

PR #165, opened 2018-08-28
added README to clairify calico and EKS

vchan2002:
Added a bit of info to clarify that EKS can also run Calico too..... I learned that the hard way myself...

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,10 +140,10 @@ variable, and running the container in a privileged security context.
 Note that the interface `--in-interface` above or using the `--host-interface` cli flag may be
 different than `docker0` depending on which virtual network you use e.g.
 
-* for Calico, use `cali+` (the interface name is something like cali1234567890
+* for Calico, use `cali+` (the interface name is something like cali1234567890)(It is possible for [EKS to use Calico also](https://docs.aws.amazon.com/eks/latest/userguide/calico.html))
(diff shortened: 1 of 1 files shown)


jrnt30:
Thanks

robinjoseph08:
@vchan2002 correct me if I'm wrong, but I believe those instructions on the AWS docs to setup Calico is only setting it up for network policies. I just installed it on a brand new EKS cluster and the network interfaces didn't change to `cali+`. It's still using `eni+` from `amazon-vpc-cni-k8s`.

jrnt30:
It's based upon the CNI provider that you choose.  The default for EKS would be the `amazon-vpc-cni-k8s` and result in `eni+` as I understand it, that is correct. 

Not exactly sure how to structure the docs so they cater to different levels of depth and understanding there.

robinjoseph08:
Ahh gotcha. I think the confusion for me stemmed from the specific link that is included (https://docs.aws.amazon.com/eks/latest/userguide/calico.html). If you follow the instructions at the link, it doesn't change the underlying CNI AFAIK. It just enables the network policies, but it still uses `amazon-vpc-cni-k8s`. So maybe it could link to a page that does show how to actually use replace `amazon-vpc-cni-k8s` with Calico?

Issue #164, opened 2018-08-28
cannot override sts endpoint


pr8kerl:
We have been using kube2iam for some time now with great success - so thank you. 

However we have had a few issues relating to the timely provisioning of credentials to pods. Recently in particular we had a situation where we were seeing a number of `InvalidTokenId` errors on pods with kube2iam annotations. These errors were occurring across a couple of kubernetes clusters and across two separate aws accounts. We raised a support ticket with AWS who said there was no sts service issue at that time, however they suggested we should be using our local region (ap-southeast-2) sts endpoint instead of the default endpoint (us-east-1). When using the default sts endpoint, any credentials created there need to be replicated to the region where they will be used. Our pods appeared to be attempting to use the sts credentials before they had replicated to our region - hence the `InvalidTokenId` error.

There have been a couple of other issues relating to the sts endpoint and region such as #72 and #103 

The answer to those issues was to set the AWS_REGION environment variable. However setting the region in this way does not appear to affect the STS endpoint used which always seems to default to `sts.amazonaws.com` regardless of the use of AWS_REGION or AWS_DEFAULT_REGION vars.

I've done some local tests with the codebase and found that the only way to ensure the endpoint is set correctly is to set it explicitly with the sts session Config and to also set the region.
This behaviour seems to be alluded to [here](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html#id_credentials_temp_enable-regions_writing_code)

Would you agree that this is the current behaviour?? That the use of region env vars does not affect the sts endpoint in use.

If so, would you be open to a PR to allow the use of a cli flag to set the sts endpoint? 

Thanks in advance.

jrnt30:
I have not dug into the implementation details of the STS client specifically as closely as you have.  If you look at the CloudTrail logs that might shed some light on where, when the env vars are properly configured, the STS token is created at. 

If you're assessment is correct, that it is not respecting that Env flag, then a PR with additional configuration would seem reasonable and accepted, yes. 

Thanks for the diligence and detail!

jtblin:
> This paragraph also describes the current

Which paragraph? :)

pr8kerl:
whoops. I think I got sidetracked at that point. :)

[...]

PR #166, opened 2018-08-29
allow use of sts regional endpoints

pr8kerl:
This one addresses #164 and allows the use of regional sts api endpoints instead of the default us-east-1 based endpoint. 

For users in regions other than us-east-1 this should mean slightly reduced latency in sts api calls.

By default the aws sdk will resolve the sts endpoint to `sts.amazonaws.com` as described [here](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp.html#sts-regionalization). The regional api endpoint can only be set programmatically. Setting region via code or environment does not affect the endpoint, as indirectly described [here](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html#id_credentials_temp_enable-regions_writing_code).

To enable the use of this feature, a boolean flag `--use-regional-sts-endpoint` must be used in conjunction with the appropriately set environment variable `AWS_REGION`.  The use of a flag will mean no impact to existing users' deployments.


SharpEdgeMarshall:
seems good to me

PR #171, opened 2018-10-03
Change ADD to COPY

dragoonis:
ADD is reserved for pulling in and automatically unpacking .tar/.gz files
COPY is for copying files across, which is what you're trying to achieve.


SharpEdgeMarshall:
lgtm

PR #177, opened 2018-11-08
Add support for sts external-id authentication

ElliotGardiner:
# Problem Statement
Currently IAM Roles created with an External ID cannot be proxied by kube2iam. Including an External ID is a [recommended design pattern when sharing IAM Roles across AWS accounts](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html#external-id-purpose). External IDs should be optionally configurable in pod metadata to support this type of IAM Role assumption.

# Summary
This change adds support for the `--external-id` parameter of `aws sts assume-role` [api calls](https://docs.aws.amazon.com/cli/latest/reference/sts/assume-role.html).

With this change clients can create pods with an external-id in addition to the role annotation. For example, the following IAM Role (`arn:aws:iam::123456789011:role/an-iam-role-with-an-external-id`)'s trust relationship document created in Account A (123456789011) allows Account B (123456789012) that hosts the cluster to assume a role that grants access to Account A resources:
```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "AWS": "arn:aws:iam::123456789012:root"
      },
      "Action": "sts:AssumeRole",
      "Condition": {
        "StringEquals": {
          "sts:ExternalId": "external-id-preventing-confused-deputy-syndrome"
        }
      }
    }
  ]
}
```
Given that trust relationship document, pods can now be registered to a cluster in Account B with the following pod yaml:
```
apiVersion: v1
kind: Pod
metadata:
  name: a-pod-with-an-external-id
  labels:
    name: a-pod-with-an-external-id
  annotations:
    iam.amazonaws.com/role: arn:aws:iam::123456789011:role/an-iam-role-with-an-external-id
    iam.amazonaws.com/external-id: external-id-preventing-confused-deputy-syndrome
```

### DIFF
--- a/iam/iam.go
+++ b/iam/iam.go
@@ -70,9 +70,9 @@ func GetInstanceIAMRole() (string, error) {
 	return iamRole, nil
 }
 
-func sessionName(roleARN, remoteIP string) string {
+func sessionName(roleARN, externalID string, remoteIP string) string {
 	idx := strings.LastIndex(roleARN, "/")
-	name := fmt.Sprintf("%s-%s", getHash(remoteIP), roleARN[idx+1:])
+	name := fmt.Sprintf("%s-%s-%s", getHash(remoteIP), roleARN[idx+1:], getHash(externalID))
 	return fmt.Sprintf("%.[2]*[1]s", name, maxSessNameLength)
 }
 
@@ -128,7 +128,8 @@ func (iam *Client) EndpointFor(service, region string, optFns ...func(*endpoints
 // AssumeRole returns an IAM role Credentials using AWS STS.
 func (iam *Client) AssumeRole(roleARN, externalID string, remoteIP string, sessionTTL time.Duration) (*Credentials, error) {
 	hitCache := true
-	item, err := cache.Fetch(roleARN, sessionTTL, func() (interface{}, error) {
+	roleCacheKey := roleARN + ":" + externalID
+	item, err := cache.Fetch(roleCacheKey, sessionTTL, func() (interface{}, error) {
 		hitCache = false
 
 		// Set up a prometheus timer to track the AWS request duration. It stores the timer value when
@@ -152,7 +153,7 @@ func (iam *Client) AssumeRole(roleARN, externalID string, remoteIP string, sessi
 		assumeRoleInput := &sts.AssumeRoleInput{
 			DurationSeconds: aws.Int64(int64(sessionTTL.Seconds() * 2)),
 			RoleArn:         aws.String(roleARN),
-			RoleSessionName: aws.String(sessionName(roleARN, remoteIP)),
+			RoleSessionName: aws.String(sessionName(roleARN, externalID, remoteIP)),
 		}
 		if externalID != "" {
 			assumeRoleInput.ExternalId = aws.String(externalID)
@@ -174,7 +175,7 @@ func (iam *Client) AssumeRole(roleARN, externalID string, remoteIP string, sessi
 		}, nil
 	})
 	if hitCache {
-		metrics.IamCacheHitCount.WithLabelValues(roleARN).Inc()
+		metrics.IamCacheHitCount.WithLabelValues(roleCacheKey).Inc()
 	}
 	if err != nil {
 		return nil, err
--- a/iam/iam.go
+++ b/iam/iam.go
@@ -126,7 +126,7 @@ func (iam *Client) EndpointFor(service, region string, optFns ...func(*endpoints
 }
 
 // AssumeRole returns an IAM role Credentials using AWS STS.
-func (iam *Client) AssumeRole(roleARN, externalId string, remoteIP string, sessionTTL time.Duration) (*Credentials, error) {
+func (iam *Client) AssumeRole(roleARN, externalID string, remoteIP string, sessionTTL time.Duration) (*Credentials, error) {
 	hitCache := true
 	item, err := cache.Fetch(roleARN, sessionTTL, func() (interface{}, error) {
 		hitCache = false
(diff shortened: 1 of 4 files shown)


mappings/mapper.go
```diff
@@ -86,7 +110,7 @@ func (r *RoleMapper) checkRoleForNamespace(roleArn string, namespace string) boo
 
 	ns, err := r.store.NamespaceByName(namespace)
 	if err != nil {
-		log.Debug("Unable to find an indexed namespace of %s", namespace)
+		log.Debugf("Unable to find an indexed namespace of %s", namespace)
```
ElliotGardiner: This change wasn't specific to external-id, but was breaking `make test`.

mappings/mapper.go
```diff
@@ -51,9 +53,19 @@ func (r *RoleMapper) GetRoleMapping(IP string) (*RoleMappingResult, error) {
 		return nil, err
 	}
 
-	// Determine if normalized role is allowed to be used in pod's namespace
+	externalID, err := r.extractExternalID(pod)
```
jrnt30: Are there any errors that could be returned? If not, this can be simplified

ElliotGardiner: Understood, not currently. I did it like this to be consistent with RoleMapper.extractRoleARN, I can strip out the error passing/checking to simplify.

jrnt30:
Need to reorient myself with some of the caching code, but in the past one thing people have overlooked when attempting to add in some "unique" attributes for credentials is the fact that right now they are cached by the role.  If the idea is that each discrete set of Role/ExternalId should be unique and specific we need to ensure that the additional ExternalId is part of the key.

Apologies but don't have time today to do that validation but wanted to give you a heads up on what I would be looking at if you have time to address sooner.

ElliotGardiner:
@jrnt30 interesting - thanks for the heads up. I can imagine a scenario where a pod is created with just a role then subsequently updated with an external id. At that point the cache should be invalidated and credentials should be re-fetched. I'll take some time today to explore the caching logic and see if there's an opportunity to add additional tests.

iam/iam.go
```diff
@@ -126,7 +126,7 @@ func (iam *Client) EndpointFor(service, region string, optFns ...func(*endpoints
 }
 
 // AssumeRole returns an IAM role Credentials using AWS STS.
-func (iam *Client) AssumeRole(roleARN, remoteIP string, sessionTTL time.Duration) (*Credentials, error) {
+func (iam *Client) AssumeRole(roleARN, externalID string, remoteIP string, sessionTTL time.Duration) (*Credentials, error) {
 	hitCache := true
 	item, err := cache.Fetch(roleARN, sessionTTL, func() (interface{}, error) {
```
ElliotGardiner: It looks like this is the crux of the caching logic. For this `ccache` get-set mechanism, updating this key to `roleARN:externalID` would help with cache invalidation when an external id is added to an iam role. However it looks like there's no mechanism for invalidating the cache if a role's credentials are invalidated before it's original TTL (invalidation seems like a broader issue).

One other point I'll add, if an iam role requires an external id, there shouldn't be a scenario where two pods are successfully using that role but only one of them is passing in the external id. If Pod A isn't using the external id, sts.AssumeRole requests will fail and nil'ify the cache https://github.com/jtblin/kube2iam/pull/177/files#diff-d40ff635e38cff8f5959411f50e5a8a6R163
meanwhile Pod B is using the external id and is populating the cache with healthy credentials that Pod A could then grab.

TLDRl; I need to update the key to include external id.

jrnt30: yeah, I am not sure how you intend to use the external id field, however it would still be possible if two pods with the same role/external id to be co-located on the same host, which would result in a "cache hit" as well.

If you are trying to get "Container Identity" built into the STS role assumption so you know *exactly* which pod is making the request (which could be very valuable in some compliance settings) this still won't address the issue.

If you could outline the "problem statement" of the PR that you are hoping to address that may be helpful ensuring we are aligned on the implications of the changes and how we might architect kube2iam to accommodate those needs.

ElliotGardiner: >it would still be possible if two pods with the same role/external id to be co-located on the same host, which would result in a "cache hit" as well.

I think that type of cache hit is acceptable for my use case (for now).

>If you are trying to get "Container Identity" built into the STS role assumption so you know exactly which pod is making the request (which could be very valuable in some compliance settings) this still won't address the issue.

Understood - that's not part of my use case (yet) - but I can see the compliance oriented value. I imagine beyond adding pod name to the cache key could also offer another solution where kube2iam prevents a role from being associated with a pod more than once (or scope it to more than once per namespace and add pod namespace to the cache key). In other words, adding a configuration flag to kube2iam that enforces a 1:1 relationship between either pod:iam-role or namespaced-pod:iam-role.

>If you could outline the "problem statement" of the PR that you are hoping to address that may be helpful ensuring we are aligned on the implications of the changes and how we might architect kube2iam to accommodate those needs.

Will do.

ElliotGardiner: Marking this key uniqueness conversation as resolved for now.

mappings/mapper.go
ElliotGardiner: This has been simplified.

ElliotGardiner:
@jrnt30 pushed some updates last night based on our conversations so far that I think addresses some of our primary concerns. Please let me know if there's anything else I can do.

mtougeron:
@jrnt30 / @ElliotGardiner any chance this can be wrapped up? This would be a huge help to my team.

Jacobious52:
Looks like the `external-id` functionality has been added in #213

jtblin:
Closing as external-id support was added.

Issue #178, opened 2018-11-20
Intervace conversion fails, causing pods not to get credentials.


kinghajj:
Sometimes, pods in my cluster fail with errors like "fatal error: Unable to locate credentials". Checking the kube2iam logs for the instance on the same node, I see entries like this:

```
time="2018-11-20T18:40:26Z" level=error msg="PANIC error processing request: interface conversion: interface {} is nil, not *v1.Pod" req.method=GET req.path=/latest/meta-data/iam/security-credentials/ req.remote=172.19.97.48 res.status=500
```

This occurs on an EKS cluster running Kubernetes 1.10.3.

tmedford:
We have seen this error as well. We upvoted to have this pulled in as well.

admssa:
getting this on EKS 1.11.5

jrnt30:
I will look at updating tomorrow

jrnt30:
You only started seeing this is 0.10.6?

admssa:
@jrnt30 Yes, I used "latest" tag for docker image. this only affected the pods on the nodes that were created after 0.10.6 release

hgranillo:
> rolled back to 0.10.4 error is disappeared.
> i won't be using "latest"
> i won't be using "latest"
> repaet 10 times :)

Found out the same today, thanks to your comment I realized that I did the same.

lumixen:
Saw this as well in ours EKS.

thealmightygrant:
Ditto. EKS 1.10.11

cnelson:
Also seeing this on EKS `Server Version: version.Info{Major:"1", Minor:"11+", GitVersion:"v1.11.5-eks-6bad6d", GitCommit:"6bad6d9c768dc0864dab48a11653aa53b5a47043", GitTreeState:"clean", BuildDate:"2018-12-06T23:13:14Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}`, roll back to `0.10.4` fixes it.

yevhenvolchenko:
The same on AWS EKS `Server Version: version.Info{Major:"1", Minor:"10+", GitVersion:"v1.10.11-eks", GitCommit:"6bf27214b7e3e1e47dce27dcbd73ee1b27adadd0", GitTreeState:"clean", BuildDate:"2018-12-04T13:33:10Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}`

[...]

PR #179, opened 2018-11-20
Update k8s client to attempt to resolve issue #178.

kinghajj:
Based on the suggestion of @rifelpet, I updated k8s.io/client-go to tag kubernetes-1.10.10. I've been running an image with this for a few hours, and it seems not to exhibit the issue anymore.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -216,7 +216,7 @@ imports:
   - pkg/util/runtime
   - pkg/util/wait
 - name: k8s.io/client-go
-  version: e121606b0d09b2e1c467183ee46217fa85a6b672
+  version: 249fa7a1d3aa05d8e423e8de91ffedd5a58243f5
   subpackages:
   - discovery
   - kubernetes
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/jtblin/kube2iam"
+	log "github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/pkg/api/v1"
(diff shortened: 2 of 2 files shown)


jrnt30:
I cherry-picked/updated your debugging code so we can try and identify the root cause here, I'm not confident the client update will clear things up.

There is an issue with the PR however, clearing out your `~/.glide` folder and `rm -Rf vendor/` results in the build failing (see Travis link).  

I will try and get back to this soon-ish unless you are able to get this to pass.

kinghajj:
No, you're right about the problems, my original 'working' version was some kind of frankenstein of dependencies. Clearing out those folders resulted in build errors, which led me down a rabbit hole of changes to the upstream K8s libs that I didn't have time (nor feel confident) to tackle. Thanks for looking into this, though.

rifelpet:
It appears that they broke apart the client-go repo into `api` and `apimachinery` repos. I have the client libraries updated to 1.12.5 (arbitrarily chosen) and successfully compiling here: https://github.com/rifelpet/kube2iam/commits/1.12-api

I haven't had a chance to actually test it though. Is there a desired version to update to?

jrnt30:
@rifelpet this looks like a step in the right direction, thank you.  Can you try pinning to [`v10.0.0` per the Client-Go docs](https://github.com/kubernetes/client-go/blob/master/INSTALL.md#glide) and [updating Glide](https://github.com/kubernetes/client-go/blob/master/INSTALL.md#glide)

rifelpet:
Done, luckily all it took was updating glide.yaml. I pinned client-go to v10.0.0 and api and apimachinery to the latest kubernetes-1.10.X releases.

https://github.com/rifelpet/kube2iam/tree/1.10-api

jrnt30:
Moved to #188 to allow me to create an RC release off the branch

PR #186, opened 2019-01-29
Add kustomize config

mgoodness:
This adds a set of manifests for use as a [kustomize](https://github.com/kubernetes-sigs/kustomize) base layer. There might also an argument to be made for simplifying the README and pointing people to these instead.

### DIFF
--- a/kustomize/daemonset.yaml
+++ b/kustomize/daemonset.yaml
@@ -1,0 +1,40 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  labels:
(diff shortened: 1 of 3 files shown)


jrnt30:
Thanks for the contribution.  Having a few overlays for the different providers could be useful to demonstrate those config values.  Look forward to seeing where the tool/community goes.

mgoodness:
I'd be happy to submit the overlay we (Ticketmaster) use, as an example.

jrnt30:
That would be great, thanks!

PR #189, opened 2019-02-05
Updates for K8 client package

jrnt30:
* Adds in error messages for pod interface failures
* Updates K8 client-go to 1.10 version
* Adjusts travis CI docker push to not tag with latest for RC builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ install:
   - make setup
 
 script:
+  - make version
   - make build
   - make test-race
   - make check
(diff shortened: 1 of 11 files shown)


geekofalltrades:
As a bonus, this MR pulls in a version of client-go that has switched from glog to klog, so that the process doesn't just [randomly die sometimes if it can't write a file in /tmp](https://github.com/kubernetes/kubernetes/issues/61006). We hit this in our environment, where we're running kube2iam from a scratch container.

WAMarcantel:
@jtblin Can we get this merged any time soon?

angulito:
any news about merging this PR? it will fix the read-only issue with `/tmp` - thank you!

PR #191, opened 2019-02-06
Add warning regarding --iptables=true to README

### LINKED ISSUE #170: --iptables option: race condition?
I'm currently setting up kube2iam in our cluster and just thought about that. If you let kube2iam set up the iptables rule, is there not a security risk by means of a race condition? If you manage to start a pod on an new instance before kube2iam is started on this node, you can basically assume the instance role.

Is this correct or am I missing something here? If this is correct, this should be documented with a large warning that this can happen and if you don't trust the pods that get deployed to the cluster you should prefer to set up this iptables rule *before* starting the kubelet.

devkid:
This closes #170 by adding a warning about a possible race condition when using `--iptables=true` to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,8 @@ iptables \
 This rule can be added automatically by setting `--iptables=true`, setting the `HOST_IP` environment
 variable, and running the container in a privileged security context.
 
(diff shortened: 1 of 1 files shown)


jrnt30:
Thanks for your contribution!

PR #192, opened 2019-02-06
Add example overlay

mgoodness:
This is the kustomize overlay in use at Ticketmaster. It's a good example (IMHO) of how to add & patch resources to a base configuration.


jrnt30:
Thanks for the example, this is useful.  

It does seem to reinforce my thought, the `args` is unable to be merged from `base` via an `overlay` which makes "composing" different features via Kustomize less approachable.  Is that accurate?

mgoodness:
"Less approachable" is exactly right. Kustomize allows for JSON patches, which allow operations that YAML does not. An arg, for example, can be appended to the existing array like so:

```json
[
  {
    "op": "add",
    "path": "/spec/template/spec/containers/0/args/-",
    "value": "--foo=bar"
  }
]
```

But it's JSON, and still somewhat limited (only array indexes, for example).

jrnt30:
Thanks, good to know it's "possible" at least.

PR #196, opened 2019-02-16
implementing iptables retry

tinder-mattheworlove:
I've been running into an issue where, when running as a daemonset, kube2iam will fail to insert its iptables rule and immediately fatals with: "iptables: Resource temporarily unavailable."

This is likely due to many processes attempting to update the iptables ruleset at the same time. Ideally there would be a retry backoff implemented to the rule adding in kube2iam as a workaround to this race condition.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -108,7 +108,7 @@ func main() {
 	}
 
 	if s.AddIPTablesRule {
-		if s.IPTablesBackoffMaxElapsedTime != "" {
+		if s.IPTablesBackoffMaxElapsedTime != nil {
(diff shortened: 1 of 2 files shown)


tinder-mattheworlove:
I've been running into an issue where, when running as a daemonset, kube2iam will fail to insert its iptables rule and immediately fatals with: "iptables: Resource temporarily unavailable."

This is likely due to many processes attempting to update the iptables ruleset at the same time. Ideally there would be a retry backoff implemented to the rule adding in kube2iam as a workaround to this race condition.

PR #197, opened 2019-02-18
Update readme for eks

stijndehaes:
Even with calico the eni... interface name is used. See https://raw.githubusercontent.com/aws/amazon-vpc-cni-k8s/master/config/v1.3/calico.yaml, environment variable: FELIX_INTERFACEPREFIX


jrnt30:
Thanks for your first contribution!

PR #199, opened 2019-02-21
Add RollingUpdate strategy to the example

### LINKED ISSUE #190: Set UpdateStrategy in DaemonSet examples
The example DaemonSet specs should include setting the [UpdateStrategy](https://kubernetes.io/docs/tasks/manage-daemon/update-daemon-set/#daemonset-update-strategy) to `RollingUpdate`.

This (kube2iam) is relatively likely to be one of a new Kubernetes user's first experience with DaemonSets if they're setting up in AWS, and they might not be aware that unlike Deployments/etc, the default for DaemonSets is for the old spec and its options to continue running indefinitely until the  pods are killed.

It's a minor papercut, but it's easy to waste time trying to figure out kube2iam's options/effects if you don't realize that the changes you've applied to the container spec aren't actually in effect on the nodes.

SharpEdgeMarshall:
Fix #190

### DIFF
--- a/examples/eks-example.yml
+++ b/examples/eks-example.yml
@@ -39,6 +39,8 @@ spec:
   selector:
     matchLabels:
       name: kube2iam
+  updateStrategy:
+    type: RollingUpdate
   template:
     metadata:
       labels:


SharpEdgeMarshall:
@jrnt30 I'm doing some cleanup could we merge this?

PR #200, opened 2019-02-26
Add AssumeRole error cache

schleyfox:
When AssumeRole is successful, the success is cached for (by default) 15
minutes.  On failure, there is no cache.  In cases of misconfiguration,
this can result in a large number of AssumeRole calls against the AWS
API, which, in extreme cases, can result in API rate limiting, causing
other applications in the AWS account to experience failures.

This adds a negative cache on AssumeRole to prevent error cases from
spamming the API.  This is set to 0 by default to preserve existing behavior,
but a value of 30 seconds to 1 minute would be reasonable.


schleyfox:
@jrnt30 can you look at this?

jtblin:
Rebased and PR #397.

PR #173, opened 2018-10-17
Allow pods being terminated to get credentials

rifelpet:
When a pod is deleted, the containers are issued a `SIGTERM` and its `deletionTimestamp` is set to now + `terminationGracePeriodSeconds`. After that time, a `SIGKILL` is issued and the pod is deleted (see [deletionTimestamp](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#objectmeta-v1-meta)).

For pods with longer terminationGracePeriodSeconds, its possible for the container process to refresh its credentials after the `SIGTERM` but before the `SIGKILL`. Prior to this PR, credentials were not provided and the process' cleanup would fail. This PR allows containers in the process of shutting down to refresh their credentials.

Fixes #172 

This means that we're now relying on the client's clock to be in sync with the master's clock but I would assume that is already a reasonable requirement for functioning Kubernetes clusters.

To test:

```sh
#!/bin/bash
# script.sh

trap "echo got shutdown" SIGTERM

while :
do
  sleep 1
done
```

```dockerfile
FROM mesosphere/aws-cli:latest

ADD script.sh /script.sh

ENTRYPOINT ["sh", "/script.sh"]
```

Run a pod with that image and an IAM role annotation. `kubectl exec` into the pod and repeatedly run `aws sts get-caller-identity` and observe the result. `kubectl delete` the pod. The credentials will continue to be fetched until the pod is terminated.

PR #203, opened 2019-03-07
Do not check the pod's deletionTimestamp when determining if a pod is active

rifelpet:
Fixing a regression originating from #173, this was causing nil entries in the pod index which resulted in credentials failing to be issued.

See #178 for more information.

### DIFF
--- a/pod.go
+++ b/pod.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


lbernail:
Thanks a lot!

rifelpet:
@jrnt30 was involved in merging and releasing #173 - any chance you can do the same for this?

lbernail:
Quick follow-up: We've been running this for about 3 weeks without any issue and everything works fine

PR #213, opened 2019-05-17
Add support for externalId as an optional annotation

ColinHebert:
AWS supports a filtering system via `ExternalId` which allows to verify that the system attempting to assume the role is not making a mistake.
This adds the support of an externalId annotation (defaults to `iam.amazonaws.com/external-id`) so when the prod is created, an externalId can be provided and passed through as part of the `assume-role` operation.


struz:
LGTM, have looked over this with @Jacobious52

PR #218, opened 2019-07-12
Update README.md

sarah97:
updated example to use working API, otherwise kubectl returns no matches for kind "CronJob" in version "batch/v1"


SharpEdgeMarshall:
lgtm

PR #226, opened 2019-08-15
debugging: only print the user-agent

dadux:


dadux:
Sorry fat-fingered and opened accidentally  !

PR #229, opened 2019-08-27
WIP: updated mapper to support comma delimited annotation

### LINKED ISSUE #153: annotate multiple roles to be assumed
Is there a way to 
Add an iam.amazonaws.com/role annotation to the pods with multiple roles that you want to assume for this pod.

cpoole:
support multiple iam roles in the annotation.

Closes #153

Testing this now, will remove WIP once verified

### DIFF
--- a/mappings/mapper.go
+++ b/mappings/mapper.go
@@ -109,7 +109,7 @@ func (r *RoleMapper) checkRoleForNamespace(roleArn string, namespace string) boo
 
 	ns, err := r.store.NamespaceByName(namespace)
 	if err != nil {
-		log.Debug("Unable to find an indexed namespace of %s", namespace)
+		log.Debugf("Unable to find an indexed namespace of %s", namespace)
 		return false
 	}
 
--- a/mappings/mapper.go
+++ b/mappings/mapper.go
@@ -7,7 +7,7 @@ import (
 
 	glob "github.com/ryanuber/go-glob"
 	log "github.com/sirupsen/logrus"
-	"k8s.io/client-go/pkg/api/v1"
+	v1 "k8s.io/client-go/pkg/api/v1"
 
 	"github.com/jtblin/kube2iam"
 	"github.com/jtblin/kube2iam/iam"
@@ -34,7 +34,7 @@ type store interface {
 
 // RoleMappingResult represents the relevant information for a given mapping request
 type RoleMappingResult struct {
-	Role      string
+	Roles     []string
 	IP        string
 	Namespace string
 }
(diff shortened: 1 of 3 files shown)


jrnt30:
Thanks for the work here, it has been quite a while since I have actively worked on this code base (or Go) honestly.  I'm not seeing though where or how the actual proxy piece would "choose" a set of STS credentials to pass back to the client.

What set of credentials should actually be used in this case?  If the roles have divergent permissions wouldn't we need to be explicit about the set we are looking to retrieve?

jmccarty3:
@cpoole Is there any remaining work for this? Do you need additional reviews?

n0rig:
Can someone look into this PR?

jrnt30:
I added a comment above.  How do people anticipate this working when there are multiple IAM roles assigned to a pod?  This would be similar to trying to have multiple IAM roles assigned to an EC2 instance which is not supported.

I don't believe that the implemented feature here would achieve the results that the user is looking for.  When a request is made, they will get the credentials from one of the relevant IAM roles via the Metadata service.

Am I missing something on how people leverage the metadata service?  Do you have this running in an environment with your own forked build by chance?

vanga:
I have same question as @jrnt30 
 The role that we use in annotation isn't the role that will be attached to the EC2 worker node, worker node just need permissions to be able to invoke `assumeRole` for the role we give in annotation. If I have to guess, the only way I would expect this to work is that applications would have to make a choice of which role to assume of all the roles given?

aramhakobyan:
any progress on this ?

aramhakobyan:
Our use case: We have multiple aws accounts and our grafana reading accounts cloudwatch logs. Grafana supports assuming role and so it's possible to access  according dashboard of particular cloudwatch/aws-account to read the metrics.
with kube2iam ... it's not possible now ... 
Please speed up testing and release the version with support

jrnt30:
Unfortunately by the time that the request comes to Kube2iam all that we have is the IP of the requesting pod, we do not have context around the assumed role.

If the application itself is supporting role assumption, that can work but the role you impart to the pods role will need Sts perms to assume those other roles

aramhakobyan:
> If the application itself is supporting role assumption, that can work but the role you impart to the pods role will need Sts perms to assume those other roles

yes that is the case - application(grafana) supports assuming multiple roles, but in order to control this via kube2iam, we have to have support to apply multiple roles on the pod level. 
This use case becoming more and more required, since AWS pushing and all companies more and more migrating in the direction of having one central organisation managing different aws accounts per department/team - which allow better isolation, and easier access and control management.

I know at least 10+ mid size companies recently moving toward that direction.

Bottom line: if the PR is valid and there is only testing left - perhaps we can help you to run a test and validate(or even fix) PR to achieve quicker release

jrnt30:
I think you are conflating the ability of the application (grafana) to assume roles and having the metadataservice on the Ec2 instance assume multiple roles. 

In the former case (which is the only thing supported by Kube2IAM or AWS), is that the credential provided to the application (grafana) is to provide it a credential which has the ability to assume roles. 

The application then uses the equivalent of an `aws sts assume-role` programmatically to assume the destination role.  

In order to facilitate this, what you would need to do is ensure that the IAM role associated with the Pod via your annotation has the ability to assume those roles.

PR #250, opened 2019-12-30
New flag --namespace-iam-role-key and the ability to request a role at the namespace level

vladiacob:
The possibility to assign a specific role at the namespace level.
A new flag to specify under which annotation the iam role is present at the namespace level.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ _testmain.go
 /build/
 coverage.out
 /vendor/
+.idea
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -29,3 +29,4 @@ _testmain.go
 /build/
 coverage.out
 /vendor/
+.idea
```
grosser: best use a global .gitignore for that

mappings/mapper.go
```diff
@@ -47,14 +48,24 @@ func (r *RoleMapper) GetRoleMapping(IP string) (*RoleMappingResult, error) {
 		return nil, err
 	}
 
+	namespace := pod.GetNamespace()
```
grosser: only do this if the flag is enabled ?

PR #263, opened 2020-03-05
Upgrade dependencies

### LINKED ISSUE #255: Rebuild docker container to patch for CVE-2019-14697
#230 incremented the base version of Alpine from 3.7 to 3.10 to patch CVE-2019-14697

https://hub.docker.com/r/jtblin/kube2iam is still built from the Alpine 3.7 base

Would it be possible for an updated image to be built and pushed to docker hub?

mwhittington21:
Fixes #255 (once container is built)
Builds on PR#266 commits but moves to go mod as the newer versions of the k8s API work better with go mod
Updates k8s API to 1.17.3
Various build improvements (see commit message)


.gitignore
```diff
@@ -28,4 +28,3 @@ _testmain.go
 
 /build/
 coverage.out
-/vendor/
```
grosser: I usually keep this so `go mod vendor` works

struz: I don't usually use that so I missed that use case. Have reverted this and the .dockerignore, thanks.

mwhittington21: I don't usually use that so I missed that use case. Have reverted this and the .dockerignore, thanks.

PR #267, opened 2020-03-12
Use api when cache fails

ltagliamonte-dd:
Thank to the new metrics I've added last time on cacheMiss `k8s_pod_not_in_pod_cache`, i was able to monitor and discover another path when sometimes the cache doesn't contain the info about the pod/IP:
Screen Shot 2020-03-12 at 4 23 25 PM
With this PR i'm trying to fix that path as well using the API as source of truth.


ltagliamonte-dd:
closing for the moment.
the issue I'm recording seems not fixable atm because of this: 
https://github.com/kubernetes/kubernetes/issues/39113
added more details here:
https://github.com/kubernetes/kubernetes/issues/39113#issuecomment-598897344

PR #270, opened 2020-05-04
Fix #241 by removing x-forwarded-for header from request to aws

DaspawnW:
Copy solution used in KIAM also to kube2iam to fix #241

PR #279, opened 2020-09-08
:bug: Remove remote address header when any IMDSv2 tokens are used

jhuntwork:
Even after #270 we have experienced 403s from the metadata service while
using the AWS VPC CNI with any pods not using host networking. This
indicated that while the PUT requests for tokens were now working
correctly, the subsequent GET requests still had the `X-Forwarded-For`
header which AWS rejects with a 403.


mwhittington21:
Thanks for the fix!

Issue #283, opened 2020-10-19
http methods allowed


friedrich-at-adobe:
Is there any reason to have HEAD and OPTIONS enabled on route / ?

Can we disable them?

jtblin:
AFAIK the ec2 metadata API supports them. Reopen if not.

PR #284, opened 2020-10-19
Only enable GET for default route

friedrich-at-adobe:
it should fix https://github.com/jtblin/kube2iam/issues/283


jtblin:
See comment on #283

PR #291, opened 2020-11-23
[helm] Add maxUnavailable option for rolling updates

hwoarang:
Allow user to set maxUnavailable option for rolling updates in order
to speed up deployment times on large clusters.

### DIFF
--- a/charts/kube2iam/values.yaml
+++ b/charts/kube2iam/values.yaml
@@ -102,6 +102,11 @@ resources: {}
 ##
 updateStrategy: OnDelete
 
+## Maximum number of Pods that can be unavailable during the update process
+## when using updateStrategy: RollingUpdate
+## Ref: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#max-unavailable
+maxUnavailable: 1
+
 verbose: false
 
 tolerations: []
--- a/charts/kube2iam/Chart.yaml
+++ b/charts/kube2iam/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kube2iam
-version: 2.5.2
+version: 2.6.0
 appVersion: 0.10.9
 description: Provide IAM credentials to pods based on annotations.
 keywords:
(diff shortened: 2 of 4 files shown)


jtblin:
cc @mariusv @walkafwalka can you help reviewing helm charts changes pls?

mariusv (approved):
LGTM 👍

mariusv (changes requested):
Please add documentation about the `maxUnavailable` on the readme on how it will be used.

walkafwalka:
I prefer using something more flexible for updateStrategy but that would require a breaking change. We should take note of situations like this.
```
  updateStrategy: {}
  #  rollingUpdate:
  #    maxUnavailable: 1
  #  type: RollingUpdate
```

Besides the documentation, LGTM.

rdelpret:
@walkafwalka I 100% agree, I reached out to @mariusv on slack about it to propose the same before I saw this PR but didn't get an answer.

rdelpret:
Regardless of implementation, this one is urgent for us as large clusters need some pretty high timeouts set w/ helm to get through a rolling update with the hard-coded settings.

charts/kube2iam/values.yaml
```diff
@@ -102,6 +102,11 @@ resources: {}
 ##
 updateStrategy: OnDelete
 
+## Maximum number of Pods that can be unavailable during the update process
+## when using updateStrategy: RollingUpdate
+## Ref: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#max-unavailable
+maxUnavailable: 25%
```
rdelpret, line 108: if the breaking change proposed by @walkafwalka is not implemented I think this value should be optional /conditional in the template and commented out. Or it set to the default value which would be `1` for a daemonset according to https://kubernetes.io/docs/tasks/manage-daemon/update-daemon-set/. You don't want to change this value out from under people.

hwoarang:
I have updated the default value and added the missing doc. I opted not to implement the braking change at this point. However, if you think we should do it then I can do that no problem

walkafwalka (approved):

PR #389, opened 2025-04-06
Make parseRemoteAddr handle IPv6 addresses

mikkeloscar:
#### What this PR does / why we need it:

This fixes the function `parseRemoteAddr` to be compatible with both IPv4 and IPv6 addresses. This enables kube2iam to work in an IPv6 enabled cluster where the pod IPs and thereby source IP/RemoteAddr would be IPv6. Without the fix, the RemoteAddr is parsed as `""` and is not mapped to the pod making the call.

Additionally it adds the address to an error so it's easier to debug issues.

#### Which issue this PR fixes
*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*
  - fixes #

#### Checklist chart
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [ ] Chart Version bumped
- [ ] Variables are documented in the README.md

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -95,7 +95,7 @@ func (k8s *Client) PodByIP(IP string) (*v1.Pod, error) {
 
 	if len(pods) == 0 {
 		metrics.PodNotFoundInCache.Inc()
-		return nil, fmt.Errorf("pod with specificed IP not found")
+		return nil, fmt.Errorf("pod with specificed IP %q not found", IP)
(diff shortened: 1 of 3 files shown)


mikkeloscar:
Was fixed as part of https://github.com/jtblin/kube2iam/commit/f0e2944f9b6e274db0905fb0d981df41afd4818e

Issue #394, opened 2025-12-17
v0.13.0 has no arm64 architecture


camaeel:
Image for version v0.13.0 has no arm64 architecture

g-sree:
@jtblin looks like that PR is merged. need a new tag release

jtblin:
Done, fixed the CI build and was able to push `latest` and `0.14.3`:

https://hub.docker.com/repository/docker/jtblin/kube2iam/tags/latest/sha256-364f51f06ba7af97c59d784930168195464866c152794e96449fc0fb81d4f41e

https://hub.docker.com/repository/docker/jtblin/kube2iam/tags/0.14.3/sha256-364f51f06ba7af97c59d784930168195464866c152794e96449fc0fb81d4f41e

Let me know how it goes.

camaeel:
Seems to work.

PR #396, opened 2026-01-15
Use docker buildx imagetools for release tagging

### LINKED ISSUE #391: Inconsistent pushes of manifest lists to dockerhub
dockerhub's `latest` tagged image, dtd May 26, 2025 at 4:14 am, is an image manifest for a linux/amd64 image, not a manifest list for a multi-arch image having at least linux/amd64 and linux/arm64[/v8]; albeit the Makefile defaults `PLATFORMS` to `linux/arm/v7,linux/arm64/v8,linux/amd64`

Unfortunately for us, we had been using `jtblin/kube2iam:latest` in our installations of kube2iam; shame on us.  This last `latest` dockerhub tag broke us on our newly scaled out Graviton based worker-group nodes!  Albeit we are attempting to have everything use IRSA, and possibly looking to move towards Pod Identity, but we still have need for kube2iam!

Even the default value for the  chart's `image.tag` parameter, `0.13.0`, in dockerhub is a single architecture image manifest.  As an emergency change, we switched the daemonset image tag in our clusters to git-hash `eb310ab`.  We apparently do not use the Chart, and deploy the k8s objects as part of our EKS stack, but whether we do decide to use the Chart or not, we'll probably have to use one of the latest git-hash image tags or `0.11.2`. ;-(

cameronattard:
#### What this PR does / why we need it:

Replace pull/tag/push workflow with docker buildx imagetools create which manipulates registry manifests directly without transferring image layers. This enables releasing multiarch images for tagged releases (dev, versioned tags, and latest) more efficiently.

#### Which issue this PR fixes

  - fixes #391, #394

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,11 +80,7 @@ dockerx:
 	docker buildx build --push --platform $(PLATFORMS) -t $(MANIFEST_NAME):$(GIT_HASH) . $(DOCKER_BUILD_FLAGS)
 
 release-dev:
-	if [ -z "$$(docker images -q $(MANIFEST_NAME):$(GIT_HASH) 2> /dev/null)" ]; then \
-		docker pull $(MANIFEST_NAME):$(GIT_HASH); \
-	fi
-	docker tag $(MANIFEST_NAME):$(GIT_HASH) $(MANIFEST_NAME):dev
-	docker push $(MANIFEST_NAME):dev
+	docker buildx imagetools create -t $(MANIFEST_NAME):dev $(MANIFEST_NAME):$(GIT_HASH)
 
 release: check test docker
 	docker push $(IMAGE_NAME):$(GIT_HASH)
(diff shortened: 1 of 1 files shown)


robertogomez80 (approved):

Makefile
```diff
@@ -96,14 +92,9 @@ ifeq (, $(findstring -rc, $(REPO_VERSION)))
 endif
 
 release-ci:
-	if [ -z "$$(docker images -q $(MANIFEST_NAME):$(GIT_HASH) 2> /dev/null)" ]; then \
-		docker pull $(MANIFEST_NAME):$(GIT_HASH); \
-	fi
-	docker tag $(MANIFEST_NAME):$(GIT_HASH) $(MANIFEST_NAME):$(REPO_VERSION)
-	docker push $(MANIFEST_NAME):$(REPO_VERSION)
+	docker buildx imagetools create -t $(MANIFEST_NAME):$(REPO_VERSION) $(MANIFEST_NAME):$(GIT_HASH)
```
jtblin, line 95: Is that going to push the image? Image is already created using `buildx` in the `dockerx` step.

cameronattard: Yes - it should push the built manifest list from `$(MANIFEST_NAME):$(GIT_HASH)` to `$(MANIFEST_NAME):$(REPO_VERSION)`.

https://docs.docker.com/reference/cli/docker/buildx/imagetools/create/

cameronattard:
Can someone please cut a new release for this?

